PR #1069, opened 2015-01-07
Fix project unsubscribes

adrianob:


app/models/user.rb
```diff
@@ -197,7 +197,7 @@ def posts_subscription
   end
 
   def project_unsubscribes
-    contributed_projects.map do |p|
+    contributions.with_state('confirmed', 'requested_refund', 'refunded').map do |p|
```
diogob: I believe we should change the meaning of ```contributed_projects``` instead of using the ```contributions.with_state('confirmed', 'requested_refund', 'refunded')```.

Take a look at https://github.com/adrianob/catarse/blob/fix_projec_unsubscribes/app/models/user.rb#L45

PR #1070, opened 2015-01-09
Dash fixes

adrianob:


app/assets/stylesheets/catarse_bootstrap/_extra.sass
```diff
@@ -27,6 +27,9 @@ select.date
 .deliver-at-date
   width: 230px
 
+.project_preview_tab
+  float: right
```
diogob: Why? Shouldn't this be in the grid?

app/controllers/admin/projects_controller.rb
```diff
@@ -8,7 +8,7 @@ class Admin::ProjectsController < Admin::BaseController
     @total_projects = Project.count(:all)
   end
 
-  [:approve, :reject, :push_to_draft, :push_to_trash].each do |name|
+  [:approve, :reject, :push_to_draft, :push_to_trash, :put_online].each do |name|
```
diogob: I think it would be better to keep the ```push_to...``` naming, like ```push_to_online```

app/controllers/projects_controller.rb
```diff
@@ -70,12 +70,21 @@ def update
     update! do |format|
       format.html do
         if resource.errors.present?
-          flash[:alert] = resource.errors.full_messages.to_sentence
+          error_message = ''
+          resource.errors.each do |error|
+            error_message += resource.errors[error][0]
+            error_message += '<br/>'
+          end
```
diogob: We should move this method to a separate class that we can use to decorate errors.

app/controllers/projects_controller.rb
```diff
@@ -70,12 +70,21 @@ def update
...
+          flash[:alert] = error_message.html_safe
         else
           flash[:notice] = t('project.update.success')
         end
 
-        redirect_to edit_project_path(@project, anchor: 'basics')
+        if params[:anchor]
```
diogob: I believe that the anchor will not be in params, besides, we should have a spec that cover the first block of the if. Probably only the else is being executed.

adrianob: The anchor is being sent by a hidden tag in the form.

app/assets/stylesheets/catarse_bootstrap/_extra.sass
adrianob: Our tab js is not working properly with tabs in separate rows, so I had to put it in the same row and just change the style.

diogob: I know the reason, let's fix the JS then, it should be more generic.

PR #1072, opened 2015-01-09
Project dashboard

devton:
- Created templates for project status on dashboard-home


app/assets/stylesheets/catarse_bootstrap/_extra.sass
```diff
@@ -98,3 +98,11 @@ select.date
 
 .hidden
   display: none
+
+.text-color-for-status
```
diogob: We should talk to @thiagocatarse and include it in the webflow

diogob:
Let's move the strings to the i18n files.

app/assets/stylesheets/catarse_bootstrap/_extra.sass
thiagocatarse: yep @diogob actually these classes were already there

rejected & failed                            -->     .text-error
waiting_funds                                -->     .text-waiting
approved, online e succesfull       -->      .text-success

maybe @devton didn't know about that

devton:
@diogob @adrianob this PR is not ready yet :P

app/assets/javascripts/app/project/edit.js
```diff
@@ -29,3 +35,40 @@ App.addChild('ProjectEdit', _.extend({
   }
 }, Skull.Tabs));
 
+// FIXME: MORE DRY HERE !!
```
devton: @diogob @adrianob i need some help here :/ I don't wanna repeat this code some suggestions?

diogob: I believe we should use the ProjectEdit for everything. Just call all the this.route

devton: @diogob the problem is the data-action generated when we use render :edit :/
or i just cant change the el to a wrapper class?

diogob: Let's use another wrapper element that you can add to the edit template and use the same for all these actions ;)

PR #1073, opened 2015-01-12
Mergemaster

adrianob:


diogob:
Problems with specs:
 * Let's remove the spec Projects::ContributionsController GET new when browser is mobile should redirect to "http://beta.catarse.me/projects/446/contributions/new" for we are not using this redirection in the bootstrap branch.
 * The specs for update_user_billing_info probably need some update. This should have been done in the current catarse/master

PR #1079, opened 2015-01-13
Home basics

devton:


app/views/catarse_bootstrap/projects/index.html.slim
```diff
@@ -0,0 +1,92 @@
...
+    .w-container
+      .w-row.u-marginbottom-20
+        .w-col.w-col-10.w-col-small-6.w-col-tiny-6
+          .fontsize-large.lineheight-looser
+            span.fa.fa-heart.text-success>.
+            | &nbsp;Próximos a você
+        .w-col.w-col-2.w-col-small-6.w-col-tiny-6
+          = link_to 'Ver todos', explore_path(ref: 'home_near', anchor: "near_of/" ) + current_user.address_state, class: 'btn btn-small btn-terciary'
+      .w-row
+        - @projects_near.each do |project|
```
diogob: Why don't we use the render_projects helper ?
https://github.com/catarse/catarse/blob/bootstrap/app/controllers/application_controller.rb#L29

PR #1092, opened 2015-01-20
Project validation

adrianob:


app/models/project.rb
```diff
@@ -135,9 +135,9 @@ class Project < ActiveRecord::Base
 
   attr_accessor :accepted_terms
 
-  # Draft state validtions
+  # Draft state validations
   validates_acceptance_of :accepted_terms, on: :create
-  validates_presence_of :name, :user, :category, :permalink
+  validates_presence_of :name, :user, :category
```
diogob: Why we are removing the permalink validation?

adrianob: We will only require a permalink when the project goes to analysis now, since we are not using the permalink anymore for the edit path.

PR #1104, opened 2015-01-27
User public profile and extras

devton:


diogob:
@devton take a look at the build

PR #1109, opened 2015-01-29
add audited account table

adrianob:


app/models/audited_account.rb
```diff
@@ -0,0 +1,7 @@
+class AuditedAccount < ActiveRecord::Base
```
diogob: Don't you think it's better to call this model ProjectBankAccount ?

devton: @diogob I think that ProjectAccount is better not? because we handle with address data too :)

app/observers/project_observer.rb
```diff
@@ -66,6 +66,29 @@ def from_approved_to_online(project)
       audited_user_moip_login: project.user.moip_login,
       audited_user_phone_number: project.user.phone_number
     })
+    AuditedAccount.create(
```
devton: @adrianob I think that we don't need this, you should use the form of dashboard_user_settings to save / create the audited_account though accept_nested_attributes... what u think?

app/models/audited_account.rb
diogob: true, ProjectAccount is better.

PR #1121, opened 2015-02-03
don't allow rewards deliver at to be in the past

adrianob:


diogob:
There are some broken specs for rewards_controller

PR #1130, opened 2015-02-05
BYE CHANNEL! hello user permalink <3

devton:


app/models/user.rb
```diff
@@ -17,7 +17,8 @@ class User < ActiveRecord::Base
     :medium_name, :display_credits, :display_total_of_contributions, :contributions_text,
     :twitter_link, :gravatar_url, :display_bank_account, :display_bank_account_owner, to: :decorator
 
-  attr_accessible :email, :password, :password_confirmation, :remember_me, :name,
+  # FIXME: Please bitch...
```
josemarluedke: :+1:

josemarluedke:
oa6lhik

app/models/user.rb
devton: :heart:

app/models/project.rb
```diff
@@ -290,10 +275,6 @@ def notify_to_backoffice(template_name, options = {}, backoffice_user = User.fin
     end
   end
 
-  def has_blank_service_fee?
```
adrianob: This shouldn't be removed.

devton: added back, weird because I don't remember to remove this method :/

lib/subdomain_constraint.rb
```diff
@@ -1,5 +1,5 @@
 class SubdomainConstraint
   def self.matches?(request)
-    Channel.pluck(:permalink).include? request.subdomain
+    User.with_permalink.pluck(:permalink).include? request.subdomain
```
diogob: And if we were 1M ?

diogob: Seriously, I think we should just test for some reserverd domains in a constant, if it's not one of the reserved domains we redirect to user#show and let the controller find the user or trigger a 404.
Otherwise this pluck is going to become very memory hungry with time.

PR #1141, opened 2015-02-11
Retina display support to README shields

marceloboeira:


diogob:
Thanks for the PR man

PR #1145, opened 2015-02-11
Climating... <3

devton:


diogob:
@devton had to undo this with a push -f, let's break it down in several PRs and discuss some things

PR #1146, opened 2015-02-11
Project account

adrianob:


app/views/catarse_bootstrap/projects/_dashboard_user_settings.html.slim
```diff
@@ -65,40 +67,38 @@
...
-                  = bform.label :owner_document, class: 'field-label fontweight-semibold force-text-dark'
-                  = bform.input_field :owner_document, as: :tel, class: 'positive'
-                .w-col.w-col-6
-    
+            .fontsize-base.fontweight-semibold= t('.account_title')
+            .fontsize-small.u-marginbottom-20= t('.account_subtitle')
+            .divider.u-marginbottom-20
+            .w-row
+              .w-col.w-col-6.w-sub-col
+                = form.label :bank_id, class: 'field-label fontweight-semibold force-text-dark'
```
devton: @adrianob try generate the form fields using form.input with wrappers because if we use input_field e can't track error style on field.

app/models/project_account.rb
```diff
@@ -0,0 +1,11 @@
+class ProjectAccount < ActiveRecord::Base
+  belongs_to :project
+  belongs_to :bank
+  belongs_to :user
+
+  delegate   :display_bank_account, :display_bank_account_owner, :display_address, to: :decorator
```
devton: you can remove this ;)
just use record.decorate.method_name on view, if we use delegate we need to delegate every new method on decorators.

PR #1150, opened 2015-02-12
Refactoring/01 - drop and refactoring some methods

devton:


app/workers/project_base_worker.rb
```diff
@@ -0,0 +1,7 @@
+class ProjectBaseWorker
+  def resource_action id, action_name
+    resource = Project.find id
+    Rails.logger.info "[#{action_name} on resource_id -> #{resource.id}] - #{resource.name}"
+    resource.send(action_name)
+  end
```
diogob: here let's use:
```ruby
def resource id
  resource = Project.find id
  Rails.logger.info "[loading resource_id -> #{resource.id}] - #{resource.name}"
end
```

app/workers/campaign_finisher_worker.rb
```diff
@@ -1,10 +1,8 @@
...
+class CampaignFinisherWorker < ProjectBaseWorker
   include Sidekiq::Worker
   sidekiq_options retry: false
 
-  def perform project_id
-    resource = Project.find project_id
-    Rails.logger.info "[FINISHING PROJECT #{resource.id}] #{resource.name}"
-    resource.finish
+  def perform id
+    resource_action id, :finish
```
diogob: And here we can use as:
```ruby
def perform id
  resource(id).finish
end
```
we can avoid the send and make the code more readable

app/workers/project_scheduler_worker.rb
```diff
@@ -1,10 +1,8 @@
...
+class ProjectSchedulerWorker < ProjectBaseWorker
   include Sidekiq::Worker
   sidekiq_options retry: true
 
-  def perform project_id
-    resource = Project.find project_id
-    Rails.logger.info "[PROJECT ONLINE #{resource.id}] #{resource.name}"
-    resource.push_to_online
+  def perform id
+    resource_action id, :push_to_online
```
diogob: same thing here

PR #1152, opened 2015-02-12
Refactoring/03

devton:


app/controllers/rewards_controller.rb
```diff
@@ -1,57 +1,11 @@
...
-    authorize @reward
-    render_form
-  end
-
-  def edit
-    authorize resource
-    render_form
-  end
-
-  def update
```
adrianob: How is the limit reward checkbox going to work?

devton: he not using this routes anymore

PR #1157, opened 2015-02-16
adjustments in copy of projects/new and ctas to guides in dashboard draft

thiagocatarse:


app/views/catarse_bootstrap/errors/not_found.html.slim
```diff
@@ -1,8 +1,9 @@
 = render 'devise/shared/header'
-.row
-  .large-5.large-centered.columns
-    .box.animated.fadeIn
-      h2.text-center= t('.title')
-      hr
-      .large-11.columns.large-centered
-        p= t('.text')
+.w-section
+	.w-container.u-text-center
```
diogob: Some indentation problem here @thiagocatarse you should adjust your text editor settings to use soft tabs with 2 spaces.

thiagocatarse: true! it was only in this file (it is an old file, so maybe the editor was using old configuration). i just adjusted it.

PR #1160, opened 2015-02-17
fetch projects on explore without having to scroll down and hide categories when searching

adrianob:


app/assets/javascripts/app/explore.js
```diff
@@ -31,13 +31,17 @@ App.addChild('Explore', _.extend({
...
+    if(window.location.hash == '' && search_string == -1){
       this.toggleCategoryList();
     }
+    else if(search_string >= 0){
+      this.fetchPage();
+    }
   },
 
   toggleCategoryList: function(event) {
+    this.fetchPage();
```
diogob: If we have a fetchPage inside this method and another one in the ```else if(search_string >= 0){``` wouldn't it be better to remove the ```else if``` and call a ```fetchPage``` after the ```if```?

PR #1166, opened 2015-02-18
Fix newrelic

adrianob:


diogob:
It seems that we had some problem updating simple_form. Take a look at the Travis log

PR #1178, opened 2015-02-20
redirect to created projects tab if user has projects

adrianob:


app/assets/javascripts/lib/tabs.js
```diff
@@ -32,7 +32,8 @@ Skull.Tabs = {
   },
 
   lookAnchors: function() {
-    $anchor = this.$('#current_anchor').data('anchor');
+    $anchor = this.$('#current_anchor').data('anchor') || window.location.hash;
```
diogob: I believe that if we put all of then in the same OR expression the code will work fine:
```javascript
$anchor = this.$('#current_anchor').data('anchor') || window.location.hash || this.$('#default-tab').data('anchor');
```

PR #1197, opened 2015-02-25
admin can see and edit data on basics, user_settings tab

devton:


app/views/catarse_bootstrap/projects/edit.html.slim
```diff
@@ -17,7 +17,7 @@
         div class="w-col #{(@project.already_deployed?) ? 'w-col-12' : 'w-col-9'}"
           = link_to '#home', id: 'dashboard_home_link', class: 'dashboard-nav-link ' , data: {target: '#dashboard_home', tab_group: 'edit'} do
             = t('.start_tab')
-          - unless @project.already_deployed? || @project.approved?
+          - if (!@project.already_deployed? || !@project.approved?) || (current_user && current_user.admin?)
```
adrianob: I think this should be  if (!@project.already_deployed? && !@project.approved?) || (current_user && current_user.admin?)

devton: sure ;)

PR #1227, opened 2015-03-10
use redactor editor for html fields

adrianob:


app/uploaders/redactor_rails_picture_uploader.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  process :read_dimensions
+
+  # Create different versions of your uploaded files:
+  version :thumb do
+    process :resize_to_fill => [118, 100]
+  end
+
+  version :content do
+    process :resize_to_limit => [800, 800]
```
diogob: It seems to me that our maximum width for project about is 680px

app/uploaders/redactor_rails_document_uploader.rb
```diff
@@ -0,0 +1,15 @@
+# encoding: utf-8
+class RedactorRailsDocumentUploader < CarrierWave::Uploader::Base
+  include RedactorRails::Backend::CarrierWave
+
+  # storage :fog
+  storage :file
+
+  def store_dir
+    "system/redactor_assets/documents/#{model.id}"
```
diogob: This probably won't work in heroku, isn't it better to just setup fog already?

app/assets/javascripts/redactor.js
```diff
@@ -0,0 +1,3 @@
+//= require redactor-rails
+//= require redactor-rails/langs/pt_br
+
```
diogob: We need a ```//=require_tree ./redactor-rails``` here to make it work in production

config/environments/production.rb
```diff
@@ -64,7 +64,8 @@
 
   # Precompile additional assets.
   # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-  # config.assets.precompile += %w( search.js )
+  config.assets.precompile += %w( search.js redacto
```
diogob: We forgot to remove this search.js (it was just an example)

app/assets/javascripts/redactor-rails/config.js
```diff
@@ -0,0 +1,28 @@
...
+        // application.js and application.css files and uncomment the line below:
+        // "plugins": ['fontsize', 'fontcolor', 'fontfamily', 'fullscreen', 'textdirection', 'clips'],
+        "imageUpload":"/redactor_rails/pictures?" + params,
+        "imageGetJson":"/redactor_rails/pictures",
+        "path":"/assets/redactor-rails",
+        "css":"style.css"
+      });
+}
+
+$(document).on( 'ready page:load', window.init_redactor );
```
vicnicius: To maintain consistency, shouldn't we move this file to `app/assets/javascript/lib/redactor.js`?

diogob: The problem is lib is included in application.js which is included for all catarse pages.
We want a separate file to include redactor only when needed.
I guess we could create a directory called vendor, but it might be confusing to have both vendor and lib.
Maybe move everything to a folder called optional ?

app/uploaders/redactor_rails_document_uploader.rb
adrianob: We don't actually need this file, I'll remove it.

app/assets/javascripts/redactor-rails/config.js
vicnicius: LGTM.

adrianob: The path to this file seems to be hard-coded on the redactor gem, I couldn't get it to read the configs after moving to another folder.

PR #1228, opened 2015-03-11
Assets files for download

thiagocatarse:


diogob:
Let's upload this files to our S3 bucket? If they are there just for download is better to not use the rails assets pipeline.

PR #1229, opened 2015-03-11
Brand assets page

thiagocatarse:
I think I made some mistake.
I deleted from my local the files that I sent in my first commit (885e0d5) but it looks like they are still in this PR?
Actually I just wanted to send the new files (included in my second commit - ec9f40f) 

One more thing: the first line in assets.html.slim is:
- content_for :title, t('.title')

I created the assets.pt.yml but I think there is something that I am missing. When I open the localhost/assets, the name of the page on the browser tab is <span class translation missing etc etc

anyway. we can see this tomorrow


diogob:
You did no mistake, it's only some part of your git traning missing ;)
I'll show you how to fix this later on.

PR #1230, opened 2015-03-11
disable account editing when account is valid

adrianob:


app/views/catarse_bootstrap/projects/_dashboard_user_settings.html.slim
```diff
@@ -1,3 +1,4 @@
+- disable_edit = @project.account.valid? && !@project.can_update_account?
```
diogob: I think it's better to move the ```@project.account.valid?``` condition into the ```Project#can_update_account?``` so you can write only: ```disable_edit = !@project.can_update_account?```

PR #1242, opened 2015-03-14
English translation

hackandgrow:
Hey guys, 
One of my team created automatic translations that use google translate to generate yml files.
Please write some comments.
I believe that this could benefit all Catarse community (especially because you do not provide en translations). 
Can you use locale string in all rails templates so language is not hardcoded.

Cheers,

Will
www.producerun.com


adrianob:
@hackandgrow we just went through a major redesign so some strings are still hardcoded. I'll work on moving all remaing strings to the yml files this week. A proper english translation is already planned.

PR #1244, opened 2015-03-16
avoid losing full_name information

adrianob:


diogob:
I believe the best solution for this case is use the same logic that we use in the other fields.
In the ```update_current_billing_info``` we use the full_name and in ```update_user_billing_info``` we use ```payer_name || full_name```

PR #1254, opened 2015-03-17
Fix explore filters

diogob:
Fixes filters when loading content on explore view. 
It is showing only recommended projects when I filter by a category.
Now we always reset all filters before following a route.


vicnicius:
Just noticed a small issue regarding those filters. Will this make those counters next to the categories name reflect the actual number of projects been shown for each category?

diogob:
Yes :D

PR #1258, opened 2015-03-17
fix unsubscribe link on posts notification

adrianob:


diogob:
Unfortunately we need this action to redirect the user using the anchor.
In case the user is not logged in, he goes to the authentication and we loose the anchor :\

PR #1262, opened 2015-03-18
Changes the link edition form layout and behavior

vicnicius:


app/assets/javascripts/app/user_links.js
```diff
@@ -1,18 +1,44 @@
...
-    var that = this;
-    this.$('a.add-user-link').unbind('click');
-    this.$('a.add-user-link').bind('click', function(event) {
-      that.$('a.user-links.add_fields').trigger('click');
-    });
+  addLink: function(event){
+
+    event.preventDefault();
+
+    this.$('a.user-links.add_fields').trigger('click');
```
diogob: Why have a method that just triggers a click in another element?

app/assets/javascripts/app/user_links.js
```diff
@@ -1,18 +1,44 @@
...
+
+  },
+
+  removeLink: function(event){
```
diogob: Are you sure this is necessary? 
Take a look at these docs: https://github.com/nathanvda/cocoon#simpleform
I believe cocoon handles this automatically. @devton has some experience with this.

PR #1264, opened 2015-03-18
Fixes dashboard progress bar

diogob:
Stops using display_progress method. It was used to display the progress bar but now it was mistakenly used to show progress in the dashboard and nothing else. So I just removed the method


devton:
Fuck yes

PR #1265, opened 2015-03-18
Local redactor

adrianob:


Gemfile.lock
```diff
@@ -136,24 +145,14 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.9.1)
-    compass (1.0.3)
+    compass (0.12.7)
```
diogob: We can't go back on compass, we need some mixins from the new version

PR #1266, opened 2015-03-18
Creating published projects property

diogob:
* Now we have the number of published projects a user has in mixpanel.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -1,5 +1,7 @@
 # coding: utf-8
 class Project < ActiveRecord::Base
+  PUBLISHED_STATES = ['online', 'waiting_funds', 'successful', 'failed']
+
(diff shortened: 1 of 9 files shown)


app/models/project.rb
```diff
@@ -139,10 +141,6 @@ def can_show_account_link?
     ['online', 'waiting_funds', 'successful', 'approved'].include? state
   end
 
-  def can_show_funding_period?
-    ['online', 'waiting_funds', 'successful', 'failed'].include? state
-  end
-
   def can_show_preview_link?
     ['draft', 'approved', 'rejected', 'in_analysis'].include? state
```
adrianob: This could be changed to !published? I think.

diogob: good catch!

PR #1269, opened 2015-03-19
Fix project contributions view for project owner

vicnicius:


app/assets/javascripts/app/project/project_contributions.js
```diff
@@ -19,15 +19,29 @@ App.views.Project.addChild('ProjectContributions', _.extend({
   },
 
   showContributions: function(){
+    
     var state = $('input:radio[name=contribution_state]:checked').val();
+
     if(state == 'waiting_confirmation'){
```
diogob: The page: 1 filter can stay before this if.
The else style we use is also a bit different.
It would be like:
```javascript
this.filter = {page: 1}
if(state == 'waiting_confirmation'){
  this.filter.with_state = 'waiting_confirmation';
}
else{
  this.filter.available_to_count = true;
}
```

PR #1270, opened 2015-03-19
add sorbonne notification

adrianob:


app/views/catarse_bootstrap/user_notifier/mailer/sorbonne.html.slim
```diff
@@ -0,0 +1,46 @@
...
+  br/
+p
+  | Obrigado!
+p
+  br/
+  br/
+p
+  | Equipe do Catarse
+
+= image_tag 'logo_sorbonne.jpg', height: '181', width: '400'
```
diogob: For the email we need the full URL of the image. I believe that the image_tag will generate a relative path no?

PR #1279, opened 2015-03-24
add alt-link style to project links

adrianob:


app/assets/stylesheets/catarse_bootstrap/_extra.sass
```diff
@@ -8,6 +8,15 @@ body
 .list-answer
   display: none
 
+.project-about, div.redactor-editor
+  a
+    display: inline
```
diogob: Why? inline should be the default for links everywhere.

PR #1281, opened 2015-03-24
Remove full name

adrianob:


diogob:
Nice! But it's missing  migrations to remove the attribute. Also, what are we going to do with the old data? Just remove it? Or it better to replace the name with the full_name?

diogob:
Also, I've included the full name in the partial app/views/catarse_bootstrap/user_notifier/mailer/_contribution_data.html.slim
So I believe its better to rebase and change this file before we merge.

adrianob:
@diogob I added the migration. Since we are now deprecating the user nickname I think it's better to overwrite all names with the full_name.

PR #1290, opened 2015-03-27
Quick Fixes and aditional specs 

vicnicius:


diogob:
There is something failing in the new spec

PR #1291, opened 2015-03-27
fix wrong redirect to users/:id/settings and added specs for then

devton:


app/controllers/users_controller.rb
```diff
@@ -22,7 +22,7 @@ def unsubscribe_notifications
 
   def settings
     authorize resource
-    redirect_to user_path(current_user, anchor: 'billing')
+    redirect_to edit_user_path(current_user, anchor: 'billing')
```
adrianob: Shouldn't the anchor be 'settings'?

devton: nope, new route on bootstrap :)

PR #1299, opened 2015-03-30
Account type

adrianob:


db/migrate/20150325224103_add_type_to_project_account.rb
```diff
@@ -0,0 +1,10 @@
+class AddTypeToProjectAccount < ActiveRecord::Migration
+  def change
+    add_column :project_accounts, :person_type, :text
+
+    execute <<-SQL
+      update project_accounts set person_type  = 'Pessoa Física' where state_inscription  is null; 
+      update project_accounts set person_type  = 'Pessoa Jurídica' where state_inscription  is not null;
```
devton: I think that you can put a query to change person_type to not null after execute the updates what you think?

diogob: I would change this name. Person type does not make much sense.
Besides, is completely dependent (therefore can be derived from) CPF/CNPJ.

Why we don't simply use a method to infer this from the document used?

app/views/catarse_bootstrap/projects/_dashboard_user_settings.html.slim
```diff
@@ -1,124 +1,128 @@
...
+              .card.u-radius.card-message.u-marginbottom-20
+                .fontsize-small
+                  = t('.account_edit_disabled_html', link: (link_to 'entrar em contato conosco'," #{CatarseSettings[:zendesk_url]}/account/dropboxes/#{CatarseSettings[:zendesk_dropbox_id]}", class: 'alt-link zendesk_widget' ))
+            .w-form.card.card-terciary
+              #email-form
+                .fontsize-base.fontweight-semibold= t('.user_data_title') 
+                .fontsize-small.u-marginbottom-20= t('.user_data_subtitle_html') 
+                .w-row
+                  .w-col.w-col-6
+                    = form.input_field :person_type, as: :select, collection: ['Pessoa Física', 'Pessoa Jurídica'], class: 'w-select positive account_type', include_blank: false
```
diogob: This does not need to be persisted.
I think this select field should just be used to call the changeType method in JS, but not go to the database.

PR #1301, opened 2015-03-30
Payments refactor

diogob:
* Creates Payment model
 * Moves contribution state machine to payment
 * Uses ContributionDetail view in admin and moves some scopes from Contribution
 * Refactor catarse_pagarme 
 * Refactor contribution creation


app/controllers/admin/contributions_controller.rb
```diff
@@ -30,7 +29,11 @@ def set_title
     @title = t("admin.contributions.index.title")
   end
 
+  def resource
+    Payment.find params[:id]
```
adrianob: Why is the resource a Payment and not a Contribution?

diogob: Yeah, this is a bit weird. But we are not using the contribution here anymore. The collection method now uses the ContributionDetail which is basically ```contributions JOIN payments```. I used the Payment as a resource so I would not have to proxy every Payment method through ContributionDetail.

app/decorators/contribution_decorator.rb
```diff
@@ -25,12 +25,29 @@ def display_value
...
   def display_slip_url
-    return source.slip_url if source.slip_url.present?
-    "https://www.moip.com.br/Boleto.do?id=#{source.payment_id.gsub('.', '').to_i}"
+    gateway_data = last_payment.try(:gateway_data)
+    return gateway_data["boleto_url"] if gateway_data.present?
+  end
+
+  def display_status
+    if last_payment.pending?
+      "Aguardando confirmação do pagamento"
```
vicnicius: Shouldn't we make those status translatable?

diogob: Good catch, I refactored the spec, and forgot to finish the method :P

app/models/contribution.rb
```diff
@@ -4,40 +4,28 @@ class Contribution < ActiveRecord::Base
...
   scope :anonymous, -> { where(anonymous: true) }
-  scope :credits, -> { where("credits OR lower(payment_method) = 'credits'") }
   scope :not_anonymous, -> { where(anonymous: false) }
-  scope :confirmed_today, -> { with_state('confirmed').where("contributions.confirmed_at::date = to_date(?, 'yyyy-mm-dd')", Time.now.strftime('%Y-%m-%d')) }
-  scope :confirmed_last_day, -> { with_state('confirmed').where("confirmed_at > current_timestamp - interval '1 day'") }
+  scope :confirmed_last_day, -> { where("EXISTS(SELECT true FROM payments p WHERE p.contribution_id = contributions.id AND p.state = 'paid' AND (current_timestamp - p.paid_at) < '1 day'::interval)") }
+  scope :pending, ->{
+    where("EXISTS (SELECT true FROM payments p WHERE p.contribution_id = contributions.id AND p.state = 'pending') AND NOT contributions.was_confirmed")
+  }
   scope :avaiable_to_automatic_refund, -> {
```
adrianob: Typo here.

adrianob: Typo on scope name.

app/models/contribution.rb
```diff
@@ -80,6 +67,10 @@ def can_refund?
     confirmed? && project.failed?
   end
 
+  def confirmed?
+    @confirmed ||= Contribution.where(id: self.id).pluck('contributions.is_confirmed').first
```
adrianob: Why not simply use find instead of where?

diogob:
@devton any remarks here?

PR #1304, opened 2015-03-31
Adds several useful properties to mixpanel contribution confirmation eve...

diogob:
...nt

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -239,7 +239,13 @@ class Project < ActiveRecord::Base
       total_contributions: self.total_contributions,
       pledged: self.pledged,
       project_state: self.state,
-      category: self.category.name_pt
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -239,7 +239,13 @@ def to_analytics_json
       total_contributions: self.total_contributions,
       pledged: self.pledged,
       project_state: self.state,
-      category: self.category.name_pt
+      category: self.category.name_pt,
+      project_goal: self.goal,
+      project_online_date: self.online_date,
+      project_expires_at: self.expires_at,
+      project_address_city: self.account.address_city,
```
adrianob: Won't this crash when there is no account?

diogob: Yup, I'm fixing it.

PR #1305, opened 2015-03-31
fix bug that prevent user that redirect to last page that visited after ...

devton:
...login

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,7 +70,7 @@ class ApplicationController < ActionController::Base
   end
 
   def redirect_user_back_after_login
-    if request.env['REQUEST_URI'].present? && !request.xhr? && session[:return_to].nil?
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -70,7 +70,7 @@ def after_sign_in_path_for(resource_or_scope)
   end
 
   def redirect_user_back_after_login
-    if request.env['REQUEST_URI'].present? && !request.xhr? && session[:return_to].nil?
+    if request.env['REQUEST_URI'].present? && !request.xhr?
```
diogob: won't this break the fallback cases? I believe we do not have specs for them. Send a feature spec for a fallback case (like logging in for the contribution through the get fallback route).

devton: @diogob this not break the fallback cases, because fallback is a real route, this session[:return_to].nil? are used when fallback_create was an action.

PR #1311, opened 2015-04-02
show user links on user card

adrianob:


spec/models/user_link_spec.rb
```diff
@@ -1,5 +1,22 @@
...
 
 RSpec.describe UserLink, :type => :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  let(:user){ create(:user) }
+
+  describe ".link_without_protocol" do
+    subject {UserLink.last.link_without_protocol}
+    context "when link has no protocol" do
+      before do
+        create(:user_link, link: 'foo.com')
```
diogob: You just need one before with ```create(:user_link, link: link)```
Then you define link in each context with ```let(:link) { ... }```
You can then use ```subject{ link.link_without_protocol}``` or, better yet, ```subject{ link.without_protocol}```

app/models/user_link.rb
```diff
@@ -1,3 +1,14 @@
 class UserLink < ActiveRecord::Base
+  before_save :fix_link
   belongs_to :user
+  scope :with_link, ->{ where('link IS NOT NULL') }
+
+  def link_without_protocol
```
diogob: As the class is called UserLink, I would rename this method to without_protocol to avoid repetition.

app/models/user_link.rb
```diff
@@ -1,3 +1,14 @@
...
+    self.link.sub(/^https?\:\/\//, '').sub(/^www./,'')
+  end
+
+  def fix_link
```
diogob: Let's rename this to something more explanatory, like prepend_protocol, or infer_protocol.

PR #1312, opened 2015-04-03
Iframe fixes

vicnicius:


app/assets/javascripts/lib/ui_helper.js
```diff
@@ -0,0 +1,69 @@
+Skull.UI_helper = {
```
diogob: The whole indentation here is weird. We use the same defaults for ruby and JS: 2 spaces

app/assets/javascripts/lib/ui_helper.js
```diff
@@ -0,0 +1,69 @@
...
+
+	setupResponsiveIframes: function($container){
+	    
+		this.$container = $container;
+
```
diogob: Too much empty lines, and empty line should be used to separate related lines, if we use it everywhere it just makes the code sparser.

PR #1316, opened 2015-04-06
Fixes Mixpanel Started Contribution event trigger

vicnicius:


app/assets/javascripts/app/mix_panel.js
```diff
@@ -45,6 +44,7 @@ App.addChild('MixPanel', {
     this.trackPageVisit('projects', 'new', 'Visited new project page');
     this.trackPageVisit('projects', 'index', 'Visited home');
     this.trackPageVisit('explore', 'index', 'Explored projects');
+    this.trackPageLoad('contributions', 'new', 'Started contribution');
```
diogob: The problem here is that the logic to detect if we have a reward already selected is missing

app/assets/javascripts/app/mix_panel.js
```diff
@@ -45,7 +44,7 @@ App.addChild('MixPanel', {
     this.trackPageVisit('projects', 'new', 'Visited new project page');
     this.trackPageVisit('projects', 'index', 'Visited home');
     this.trackPageVisit('explore', 'index', 'Explored projects');
-    this.trackPageLoad('contributions', 'edit', 'Selected reward');
```
diogob: ```trackContributions``` is too complex.
I believe it's better to leave this line and use the trackContributions only for the contribution/new in order to make it simpler.

app/assets/javascripts/app/mix_panel.js
```diff
@@ -96,6 +80,19 @@ App.addChild('MixPanel', {
     });
   },
 
+  trackContributions: function(){
+    var from = '',
+        self = this;
+    if(window.location.href.indexOf('reward_id') > -1){
```
diogob: Ths if can be inside the ```trackOnPage``` callback, since it's only needed when we have the page contributions/new.

PR #1321, opened 2015-04-08
Quick fixes

vicnicius:


app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
```diff
@@ -24,6 +24,9 @@ html
       meta property="og:site_name"    content=CatarseSettings[:company_name]
       meta property="fb:app_id"       content=CatarseSettings[:facebook_app_id]
       meta property="og:description"  content=(t('projects.index_header.title') + '. ' +  t('projects.index_header.subtitle'))
+    - if Rails.env.production?
+      = render 'layouts/hotjar'
+      = render 'layouts/vwo'
```
diogob: We should not use vwo and optimizely at the same time, I believe it's a waste of resources

app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
```diff
@@ -24,6 +24,9 @@ html
       meta property="og:site_name"    content=CatarseSettings[:company_name]
       meta property="fb:app_id"       content=CatarseSettings[:facebook_app_id]
       meta property="og:description"  content=(t('projects.index_header.title') + '. ' +  t('projects.index_header.subtitle'))
+    - if Rails.env.production?
```
diogob: We have a similar if further down this template, can't we render these partials there?

app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
```diff
@@ -24,6 +24,9 @@ html
...
+      = render 'layouts/hotjar'
```
diogob: Same thing for hotjar and crazyegg, let's use one tool for each purpose.

app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
diogob: Just confirmed with thiago, you can remove the optimizely script and replace it with the vwo

app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
diogob: Here we are removing the crazyegg and replacing with hotjar.

PR #1325, opened 2015-04-08
fix credit card destroy and add basic spec

adrianob:


spec/controllers/users/credit_cards_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    allow(controller).to receive(:current_user).and_return(user)
+  end
+
+  describe "DELETE destroy" do
+    before do
+      delete :destroy, user_id: user.id, id: credit_card.id, locale: :pt
+    end
+
+    context "when user is card owner" do
+      its(:status) { should == 302 }
```
diogob: it's always better to check
```ruby
it{ is_expected.to redirect_to helper_function_path }
```
So we can assert the redireaction is going to the right address

PR #1358, opened 2015-04-15
Last contribution state

devton:
Added last contribution state to admin/contributions


app/models/contribution_detail.rb
```diff
@@ -46,4 +47,17 @@ def self.between_values(start_at, ends_at)
     where("value between ? and ?", start_at, ends_at)
   end
 
+  def last_state_name
+    possible_states = TRANSITION_DATES.map do |state_at|
```
diogob: this could be a private method

PR #1366, opened 2015-04-16
Refund adjusts

devton:
- delegate payment.invalid_refund to contribution;
- deleted contribution can go to refunded;
- i18n adjusts

### DIFF
--- a/spec/controllers/admin/contributions_controller_spec.rb
+++ b/spec/controllers/admin/contributions_controller_spec.rb
@@ -39,59 +39,58 @@ RSpec.describe Admin::ContributionsController, type: :controller do
   end
 
   describe 'PUT trash' do
-    subject { payment.deleted? }
-
     before do
       allow(controller).to receive(:current_user).and_return(admin)
       put :trash, id: payment.id, locale: :pt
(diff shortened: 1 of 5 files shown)


spec/controllers/admin/contributions_controller_spec.rb
```diff
@@ -66,6 +66,21 @@
     end
   end
 
+  describe 'PUT request_refund' do
+    let(:contribution) { create(:confirmed_contribution) }
+    subject { payment.pending_refund? }
```
diogob: When we test controller is better to always use the response as the subject.

spec/controllers/admin/contributions_controller_spec.rb
```diff
@@ -66,6 +66,21 @@
...
+
+    before do
+      allow(payment).to receive(:direct_refund).and_return(true)
+      put :request_refund, id: payment.id, locale: :pt
+    end
+
+    it do
+      payment.reload
+      is_expected.to eq(true)
```
diogob: Using the response as subject this test would become explicit and we should use a description in the it :D

PR #1374, opened 2015-04-20
Changing the size of typo in reward card

thiagocatarse:


app/assets/stylesheets/catarse_bootstrap/_main.scss
```diff
@@ -143,7 +144,8 @@ p {
 .section.full-height.hello-idea {
   padding-top: 108px;
   padding-bottom: 169px;
-  background-image: url('../images/bg-ideia.jpg');
+  background-image: url('../images/bg-icone-hello-6.jpg');
```
devton: I think that this is wrong, we need to use assets path like:
background-image: url('/assets/image_name')

thiagocatarse: actually it is right this way @devton 

we are using this class just in mockups (webflow.com sites). 

by the time we approve the mockup and code it we will have to duplicate this class on the _extra.sass   (because I can't use the assets in webflow with the catarse url)

do you remember that you had to do the same for the hero-image classes?

diogob:
@devton @thiagocatarse I'm merging this as the code was already using this path. But Later we have to streamline our workflow so that this kind of PR would come with the assets and extra classes in place already.

thiagocatarse:
@diogob the problem about this idea of having the assets and extra classes in place already is that we use the webflow during the mockup and prototyping process. For example now I don't know yet if the images that I am using on the mockups will be the final ones, so there is no point to put them in our assets folder nor in the extra css. Did you get the point? So everytime I need to send a PR with modifications in our _main.scss while we are still doing mockups on different pages, there will be inevitably some "residues" in our css that relate to images still used in mockups.

diogob:
Ok, now I got it

PR #1375, opened 2015-04-21
Contribution and payment specs

vicnicius:


spec/factories.rb
```diff
@@ -46,10 +50,11 @@
     f.permalink { generate(:permalink) }
     f.name "Foo bar"
     f.password "123456"
-    f.cpf "123456"
+    f.cpf "97666238991"
     f.uploaded_image File.open("#{Rails.root}/spec/support/testimg.png")
     f.email { generate(:email) }
     f.about_html "This is Foo bar's biography."
+    f.country_id 1
```
diogob: should be ```f.association :country, factory: :country```

spec/support/feature_helpers.rb
```diff
@@ -12,6 +12,7 @@ def login
   end
 
   def current_user
+    FactoryGirl.create(:country)
```
diogob: Don't need to create the country here if it's set correctly in the user factory.

spec/features/projects_spec.rb
```diff
@@ -157,10 +157,101 @@
       sleep FeatureHelpers::TIME_TO_SLEEP
       fill_in "project_posts_attributes_0_title", with: 'Foo title'
       page.execute_script( "$('.redactor').redactor('code.set', 'Bar')" )
-      
       click_button(I18n.t('projects.dashboard_posts.publish'))
       expect(page).to have_content('Foo title')
     end
   end
 
+  describe "contribute" do
```
diogob: Nice specs :+1: 
But I believe we should have them in a separate file like ```spec/features/contributions_spec.rb```

spec/features/projects_spec.rb
```diff
@@ -157,10 +157,101 @@
...
+    let(:project){ create(:project, state: 'online') }
+    let(:successful_project){ create(:project, state: 'successful') }
+    let(:failed_project){ create(:project, state: 'failed') }
+
+    before do
```
diogob: I think we should move to the before block the lines
```ruby
5.times{ create(:reward, project: project) }
visit project_path(project)
```
Int the case where you have the ```visit project_path(successful_project)``` we can solve this with a ```let(:project){ successful_project }```inside a context.

PR #1377, opened 2015-04-21
replace iframe videos with link on email notification

adrianob:


app/decorators/project_post_decorator.rb
```diff
@@ -0,0 +1,17 @@
+class ProjectPostDecorator < Draper::Decorator
+  decorates :project_post
+  include Draper::LazyHelpers
+
+  def email_comment_html
```
diogob: We need a spec for this.

PR #1398, opened 2015-04-28
fix total_compromised to look only pending contributions on view

devton:


app/models/reward.rb
```diff
@@ -60,8 +60,8 @@ def any_sold?
     total_compromised > 0
   end
 
-  def total_compromised
-    payments.with_states(['paid', 'pending']).count("DISTINCT contributions.id")
+  def total_compromised states = %w(paid pending)
```
diogob: Let's create another method called total_pending.
Having this argument in total_compromised is confusing and can generate errors in sold_out? and any_sold?

PR #1401, opened 2015-04-29
Report adjusts

devton:
- provide a small refactor, we don't need to use anymore reports base_controller for one controller
- use project name on filename of csv / xls file


app/controllers/reports/contribution_reports_for_project_owners_controller.rb
```diff
@@ -24,6 +27,16 @@ def end_of_association_chain
             WHEN anonymous='f' THEN '#{I18n.t('no')}'
         END as "#{I18n.t('contribution_report_to_project_owner.anonymous')}"
       }).
-      where(conditions)
+      where(get_conditions)
+  end
+
+  def get_conditions
```
diogob: We should use has_scope for this.

PR #1408, opened 2015-05-04
Refactor timezone

adrianob:
Use only UTC on database, no timezone configuration needed anymore.
Timezone should only be considered on views.


app/observers/project_observer.rb
```diff
@@ -1,6 +1,12 @@
 class ProjectObserver < ActiveRecord::Observer
   observe :project
 
+  def before_save(project)
```
diogob: I think a spec for this would be nice.

PR #1417, opened 2015-05-06
validate reward minimum value when nested

adrianob:


app/controllers/projects_controller.rb
```diff
@@ -57,7 +57,7 @@ def update
 
     resource.attributes = permitted_params
 
-    if resource.save(validate: valid && should_use_validate)
+    if resource.save(validate: ( valid && should_use_validate ) || permitted_params[:rewards_attributes].present? )
```
diogob: should be 
```ruby
if resource.save(validate: should_use_validate)
```
Let's move all the logic to that method.

PR #1421, opened 2015-05-07
added project_status to contribution_details view to avoid multiple j…

devton:
…oins on project table


db/migrate/20150507141628_add_project_status_to_contribution_details.rb
```diff
@@ -0,0 +1,90 @@
...
+      pa.installment_value,
+      pa.state,
+      c.anonymous,
+      c.payer_email,
+      pa.gateway,
+      pa.gateway_id,
+      pa.gateway_fee,
+      pa.gateway_data,
+      pa.payment_method,
+      pa.state AS payment_state,
```
diogob: Let's remove the payment state.
We use only the state above

PR #1423, opened 2015-05-07
Adjust project metrics to use contributoin_details 

devton:


spec/controllers/projects/metrics_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "GET index" do
+    context "without an authenticated user" do
+      before { get :index, project_id: project.id,  locale: :pt }
+      it { is_expected.to redirect_to sign_up_path }
+    end
+
+    context "with authenticated user" do
+      let(:current_user) { create(:user) }
+      before { get :index, project_id: project.id,  locale: :pt }
```
diogob: before is always the same, no need to repeat in each context

diogob:
Beautiful! :+1:

PR #1428, opened 2015-05-11
show floored pledged value [94192252]

adrianob:


diogob:
to send to pivotal we should use: [#94192252]

PR #1434, opened 2015-05-12
First launching of Saiba Mais page

thiagocatarse:
@vicnicius I am sending this PR with the static Saiba Mais (Hello) page. (without animations and video lightbox) 

Tomorrow I will ask you some stuff about the Locales (I don't know how to point the Saibamais locales to this file). I can do the dirty job changing all the text within the file!

After I finish this I can give you this task.

ps: I am using a portuguese word for the URL. I think we should discuss that @diogob .. I know that we will most likely have different opinions hehe.. But I would like to think about it together ;)


diogob:
I believe the files named saibamais.html.slim and saibamais.pt.yml could be named about, or something equivalent in english.
Also there are other file names in portuguese that needs to be translated. Let's keep everything in english.

thiagocatarse:
It is done @diogob i am now using an english naming.. (as we spoke earlier)

@vicnicius lets chat in the next couple of days to discuss the next steps

PR #1435, opened 2015-05-12
Fix deliver validation

adrianob:


app/models/reward.rb
```diff
@@ -37,7 +37,8 @@ class Reward < ActiveRecord::Base
   after_save :expires_project_cache
 
   def deliver_at_cannot_be_in_the_past
-    self.errors.add(:deliver_at, "Previsão de entrega deve ser superior a data em que o projeto entra no ar") if self.project.online_date && self.deliver_at < self.project.online_date.beginning_of_month
+    self.errors.add(:deliver_at, "Previsão de entrega deve ser superior a data em que o projeto entra no ar") if
+      ( self.project.online_date && self.deliver_at < self.project.online_date.beginning_of_month ) ^ self.deliver_at < Time.current.beginning_of_month
```
diogob: Let's change this rule to use expires_at.
we should use:
```ruby
self.project.expires_at.present? ? self.deliver_at < self.project.expires_at.beginning_of_month : self.deliver_at < Time.current.beginning_of_month
```

PR #1438, opened 2015-05-12
added method to check if payment is invalid refund

devton:


lib/tasks/verify_pagarme_transactions.rake
```diff
@@ -75,10 +75,21 @@ task :verify_pagarme_transactions, [:start_date, :end_date]  => :environment do
...
+      payment.invalid_refund
+    end
+    payment.reload
+  end
+
   def status_ok?(payment, source)
     return true if payment.deleted? || payment.chargeback?
     case source['status']
     when 'paid', 'authorized' then
+      check_if_payment_is_invalid_refund(payment)
```
diogob: this method should have no side-effect

devton: @diogob  maybe rename it? because we need to change the payment status when he is refunded to peding_refund and notify backoffice about invalid_refund.

PR #1445, opened 2015-05-13
Fixes users_controller spec and updates notification link properly

vicnicius:


app/views/catarse_bootstrap/user_notifier/mailer/posts.html.slim
```diff
@@ -9,5 +9,5 @@ br
 |Veja o projeto: #{link_to update.project.name, project_by_slug_url(permalink: update.project.permalink)}
 br/
 br/
-| #{link_to 'Descadastrar', unsubscribe_notifications_user_url(user)} o recebimento de novidades desse projeto
+| #{link_to 'Descadastrar', edit_user_url(user, anchor: 'notifications')} o recebimento de novidades desse projeto
```
diogob: we should still use unsubscribe_notifications_user_url here to be able to authorize without loosing the anchor

PR #1447, opened 2015-05-13
Removes user locale from set_locale and avoid setting locale to anything not listed in AVAILABLE_LOCALES

diogob:
This will avoid several bugs existing in our english version.
Also avoids the problem of having to redirect the user when he lands in catarse (the one responsible for redirecting to /projects in the home page)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -62,7 +62,7 @@ class ApplicationController < ActionController::Base
   end
 
   def is_locale_available?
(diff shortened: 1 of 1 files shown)


vicnicius:
Shouldn't we remove the locale filter from our routes config? As we will always use one language only it would avoid showing it on the URL.

https://github.com/catarse/catarse/blob/master/config/routes.rb#L19

diogob:
@vicnicius the problem here is route compatibility with the past.
Even if we remove the locale from all generated routes we still will need the filter unless we want to break compatibility with past links.

PR #1449, opened 2015-05-13
Updates mixpanel user properties

vicnicius:


app/assets/javascripts/app/mix_panel.js
```diff
@@ -234,6 +234,7 @@ App.addChild('MixPanel', {
       default_options.has_contributions = (this.user.total_contributed_projects > 0);
       default_options.projects = this.user.total_created_projects;
       default_options.has_projects = (this.user.total_created_projects > 0);
+      default_options.has_online_project = this.user.has_online_project;
```
diogob: I think it would be easier if these options simply extended the this.user.
To achieve this we need the to_analytics_json to mirror all mixpanel properties we already have.

PR #1463, opened 2015-05-19
Fix delete pending payments

diogob:
Now the rule for deleteing pending payments is:
5 weekdays after creation.
It's a simpler version of our old rule.
I haven't touched the spec since our old spec covers this new rule and should pass with the new implementation.


devton:
<3

PR #1473, opened 2015-05-20
Fake validate

adrianob:
Just a proof of concept.. still need to work on front-end.


diogob:
I've been talking to @vicnicius about future changes to our form validation, and I think we should implement this functionality entirely on the client-side. The server side will have the validation in the database, so we won't be able to use it in the front-end.

PR #1475, opened 2015-05-21
Moves VWO script to header to prevent flash of old content on variati…

vicnicius:
…on test

### DIFF
--- a/app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
+++ b/app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
@@ -45,7 +45,6 @@ html
     /page specific JS files
     = yield :js
   - if Rails.env.production?
-    = render 'layouts/vwo'
     = render 'layouts/hotjar'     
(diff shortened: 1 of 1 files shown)


app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
```diff
@@ -26,7 +26,9 @@ html
       meta property="og:site_name"    content=CatarseSettings[:company_name]
       meta property="fb:app_id"       content=CatarseSettings[:facebook_app_id]
       meta property="og:description"  content=(t('projects.index_header.title') + '. ' +  t('projects.index_header.subtitle'))
-      
+      /Call VWO on header to prevent flash of unmodified content on variation tests
+      - if Rails.env.production?
```
diogob: This will render the vwo twice, is that correct?

vicnicius: It wasn't. Just fixed it.

PR #1485, opened 2015-05-26
Adds maxlength to project headline to avoid overflow in project card …

diogob:
…and be consistent with input hint [fix #94096994]

### DIFF
--- a/app/views/catarse_bootstrap/projects/_dashboard_project.html.slim
+++ b/app/views/catarse_bootstrap/projects/_dashboard_project.html.slim
@@ -82,7 +82,7 @@
             wrapper_class: 'w-row u-marginbottom-30 card card-terciary',
             label_wrapper_html: { class: 'w-col w-col-5 w-subcol' },
             text_field_wrapper_html: { class: 'w-col w-col-7' },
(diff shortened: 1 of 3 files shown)


app/views/catarse_bootstrap/projects/_dashboard_project.html.slim
```diff
@@ -82,7 +82,7 @@
             wrapper_class: 'w-row u-marginbottom-30 card card-terciary',
             label_wrapper_html: { class: 'w-col w-col-5 w-subcol' },
             text_field_wrapper_html: { class: 'w-col w-col-7' },
-            input_html: { class: 'positive', required: true },
+            input_html: { class: 'positive', required: true, maxlength: 100 },
```
adrianob: Are we going to keep the model validation at 140 chars?

diogob: Nice catch, I'll change there as well

devton: @diogob what you think about use `maxlength: Project::HEADLINE_MAXLENGTH` ?

PR #1487, opened 2015-05-26
Vwo experiment

vicnicius:
I am reverting those two commits as this changes are not allowing us to do proper user segmentation on VWO. Will study how to re-do this without side-effects.


diogob:
@vicnicius we have a conflict here

PR #1488, opened 2015-05-26
Turn 'Open Source' footer message into a link.

lmatiolis:
The idea is just to make the catarse code more accessible, which could end up bringing more contributors.

Ps: Also added a mixin to share style between footer links.

### DIFF
--- a/app/views/catarse_bootstrap/shared/_footer_big.html.slim
+++ b/app/views/catarse_bootstrap/shared/_footer_big.html.slim
@@ -21,6 +21,4 @@ footer.main-footer.main-footer-neg
   .w-container
     .footer-full-copyleft 
       = image_tag "logo-footer.png", class: "u-marginbottom-20"
-      .lineheight-loose Feito com amor | 2015 | Open source
-      
-      
+      .lineheight-loose: a.link-footer-inline[href="#{CatarseSettings[:github_url]}"] Feito com amor | #{Date.today.year} | Open source
--- a/app/assets/stylesheets/catarse_bootstrap/_extra.sass
+++ b/app/assets/stylesheets/catarse_bootstrap/_extra.sass
@@ -247,7 +247,6 @@ img, iframe
 .section-large.hello-catarse 
   background-image: url('/assets/bg-hello-catarse.jpg') !important
 
-  
 
 //search form style
 .discover-form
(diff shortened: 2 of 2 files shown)


app/views/catarse_bootstrap/shared/_footer_big.html.slim
```diff
@@ -21,6 +21,7 @@ footer.main-footer.main-footer-neg
...
     .footer-full-copyleft 
       = image_tag "logo-footer.png", class: "u-marginbottom-20"
-      .lineheight-loose Feito com amor | 2015 | Open source
-      
-      
\ No newline at end of file
+      .lineheight-loose
+        |  Feito com amor |
+        |  2015 |
+        a.link-footer-inline< href="#{CatarseSettings[:github_url]}" Open source
```
lmatiolis: I'm not used to Slim, so if there's a better to write this inline link tag with the text before it, please feel free to suggest the change and i'll do it! ;)

app/assets/stylesheets/catarse_bootstrap/_main.scss
```diff
@@ -634,15 +634,14 @@ p {
   background-color: #3f4752;
 }
 .link-footer {
+  @include link-footer-base;
```
diogob: This file is generated automatically by http://catarse.webflow.io/
So we'd rather have this PR without changes to the CSS

app/assets/stylesheets/catarse_bootstrap/catarse.css.sass
```diff
@@ -3,6 +3,8 @@
 @import catarse_bootstrap/normalize
 
 // The order below DOES matter. Extra should have priority.
+@import catarse_bootstrap/custom_mixins
```
diogob: Anything not in the main should go to the extra file, so we can better isolate the CSS generated by hand (not from webflow)

app/views/catarse_bootstrap/shared/_footer_big.html.slim
diogob: I would rather use the entire sentence as a link. So we wouldn't need to worry with the link being inline.
```slim
a.link-footer[href="#{CatarseSettings[:github_url]}"] Feito com amor | #{Date.today.year} | Open source
```

lmatiolis:
@diogob I didn't notice the webflow, sorry. I think i prefer the link only on the "Open Source" text, but i changed it as you suggested.

diogob:
Thanks for the PR :+1:

PR #1490, opened 2015-05-27
Tracks ref parameter [fix #95420904]

vicnicius:


app/assets/javascripts/app/mix_panel.js
```diff
@@ -217,11 +227,12 @@ App.addChild('MixPanel', {
     var page            = {
       'page name':          document.title,
       'url':                window.location,
-      'host':               window.location.host
+      'host':               window.location.host,
+      'from_page':          this.getUrlParameter('ref')
```
diogob: It's easier to get the session variable using some html data attribute.

PR #1501, opened 2015-05-28
Fix video hello mobile issues

vicnicius:


app/assets/javascripts/app/hello.js
```diff
@@ -35,12 +35,18 @@ App.addChild('Hello', {
 
   createPlayer: function(){   
     var that = this;
+    var ua = navigator.userAgent.toLowerCase();
+    var isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
```
diogob: This should be a method in the root view (App) so we could call anywhere:
```javascript
app.isMobile()
```

app/assets/javascripts/app/hello.js
```diff
@@ -35,12 +35,18 @@ App.addChild('Hello', {
...
+    if(isMobile){
+      video_crls = 1;
+    }
     that.player = new YT.Player('player', {
       height: '720',
       width: '1280',
       videoId: 'znj9ujQnrw4',
       playerVars:{
-        controls: 0,
+        controls: video_crls,
```
diogob: I'd rather have something like:
```javascript
controls: app.isMobile()
```

or

```javascript
controls: app.isMobile() ? 1 : 0
```

app/views/catarse_bootstrap/pages/hello.html.slim
```diff
@@ -3,13 +3,13 @@
 
 .hero-hello
   .w-embed.w-hidden-medium.w-hidden-small.w-hidden-tiny
-    video autoplay="" loop="" style=("width:100%; height: auto; position:absolute; z-index: -1;") 
+    video poster="/assets/hero-hello.jpg" autoplay="" loop="" style=("width:100%; height: auto; position:absolute; z-index: -1;")
```
diogob: Should use image_path helper function:
http://apidock.com/rails/ActionView/Helpers/AssetUrlHelper/image_path

PR #1502, opened 2015-05-29
Moves VWO script to HEAD fixing user segmentation problem

vicnicius:


app/views/layouts/_vwo.html.slim
```diff
@@ -1,10 +1,6 @@
 - if Rails.env.production? && CatarseSettings['vwo_id'].present?
 javascript:
-  var project = $('#project-header').data('stats'),
-      is_on_online_page = false; 
-  if(project && project.project_state === 'online'){
-    is_on_online_page = true;
-  }
+  is_on_online_page = #{defined?(@project) && @project.online? ? 'true' : 'false'}
```
diogob: Why move this logic from the JS to the backend?

vicnicius: VWO needs this var to be set before loading the experiment and as we are moving the script to the end of the Head tag, it would be simpler to set it using back-end logic as it wouldn't make sense to wait for the DOM to load.

PR #1504, opened 2015-05-29
Comments canonical url

devton:


app/controllers/projects_controller.rb
```diff
@@ -162,4 +164,10 @@ def permitted_params
   def resource
     @project ||= (params[:permalink].present? ? Project.by_permalink(params[:permalink]).first! : Project.find(params[:id]))
   end
+
+  def project_comments_canonical_url project
```
diogob: Why pass the project as parameter, it seems we will always want the project returned by ```resource```

PR #1507, opened 2015-06-02
Improves pg_search to include trigram search on Project [Fix #94112928]

lerrua:


spec/models/project_spec.rb
```diff
@@ -385,14 +385,26 @@
...
+      subject{ Project.search_tsearch('lorem') }
+      it{ is_expected.to eq([]) }
+    end
+  end
+
+  describe "#search_trm" do
+    before { @p = create(:project, name: 'criptorave') }
+    context "when project exists" do
+      subject{ [Project.search_trm('cripto'), Project.search_trm('críptõ')] }
+      it{ is_expected.to eq([[@p],[@p]]) }
```
diogob: Let's use match_array here:
```ruby
it{ is_expected.to match_array([[@p],[@p]]) }
```

app/models/project.rb
```diff
@@ -51,6 +51,15 @@ class Project < ActiveRecord::Base
     },
     ignoring: :accents
 
+  pg_search_scope :search_trm,
+    against: "name",
+    using: :trigram,
+    ignoring: :accents
+
+  def self.pg_search term
```
diogob: Let's create a spec for this. We can mock the return of search_tsearch(term) and have a context for the empty array and a context for an array containing one item.

spec/models/project_spec.rb
```diff
@@ -385,18 +385,42 @@
...
 
+  describe "#pg_search" do
+    before { @p = create(:project, name: 'criptorave') }
+
+    context "when search_tsearch is empty" do
+      before do
+        allow(Project).to receive(:search_tsearch).and_return([])
+        expect(Project).to receive(:search_trm).and_call_original
+      end
+      it { expect(Project.search_trm('cripto')).to match_array([@p]) }
```
diogob: should use the pg_search instead of search_trm

PR #1519, opened 2015-06-05
Mithril

diogob:
* Adds Mithril to bower dependencies
 * Adds database structures to authenticate API
 * Create system to keep catarse and API users in sync
 * Get JWT on client-side


db/migrate/20150605192639_add_postgrest_auth_table.rb
```diff
@@ -0,0 +1,36 @@
+class AddPostgrestAuthTable < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+      CREATE SCHEMA postgrest;
```
devton: we i'll use the schema to handle with api business logic and handle with api access on schema 1?

PR #1524, opened 2015-06-10
Clean up README, clarify instructions, prevent heroku migration error

Haseeb-Qureshi:
Cleaned up some typos on the README and added instructions for installing Bower.

Also added a line on config/environments/production.rb to prevent the Heroku migration error from popping up (see: http://stackoverflow.com/a/17306077)

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -76,6 +76,9 @@ Catarse::Application.configure do
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Prevents schema dump from throwing error in production.
+  config.active_record.schema_format = :ruby
+
   # Disable automatic flushing of the log to improve performance.
   # config.autoflush_log = false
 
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -76,6 +76,9 @@
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Prevents schema dump from throwing error in production.
+  config.active_record.schema_format = :ruby
```
diogob: We cannot accept this. The database of catarse today has several features not supported by the schema.rb
The rake db:migrate should work in heroku and the error it usually will usually report is safe to ignore:
```
rake aborted!
Error dumping database
```

diogob:
Thanks for the PR man, but we really need the schema_format to be SQL, unless you are able to make the schema.rb support the PostgreSQL functions, views and extensions. This might not be so hard since we already have the schema_plus gem, but it's not a trivial work.

diogob:
Also, the README changes can go on a PR of its own ;)

Haseeb-Qureshi:
Ah, okay. I'll submit a new pull request. :)

I'm also doing some translating work tidying up the language in en.yml, I'll submit another request probably before the end of the day with my changes.

PR #1529, opened 2015-06-12
Check if payment has changed on gateway before move to trash

devton:


app/models/payment_transfer.rb
```diff
@@ -2,4 +2,9 @@ class PaymentTransfer < ActiveRecord::Base
   # this user is the admin that authorized the transfer
   belongs_to :user
   belongs_to :payment
+
+  # retun the transfers that are pending to transfer by pagar.me
+  scope :pendings, -> do
```
diogob: there's no "s" in the plural form of pending.
should be 
```ruby
scope :pending, -> do
```

lib/tasks/cron.rake
```diff
@@ -3,7 +3,10 @@ namespace :cron do
   task hourly: [:finish_projects, :cancel_expired_waiting_confirmation_contributions, :refresh_materialized_views]
 
   desc "Tasks that should run daily"
-  task daily: [:refresh_materialized_views, :update_payment_service_fee, :notify_project_owner_about_new_confirmed_contributions, :deliver_projects_of_week, :verify_pagarme_transactions]
```
diogob: :+1:

PR #1535, opened 2015-06-22
New flow to request refund for paid payments on failed projects

devton:


app/controllers/bank_accounts_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  def permitted_params
+    params.require(:bank_account).permit(policy(resource).permitted_attributes)
+  end
+
+  def find_bank_account
+    (BankAccount.find(params[:id]) if params[:id]) ||
+      current_user.bank_account ||
```
diogob: Why use the current_user.bank_account?
Wouldn't it be better tu just use:
```ruby
params[:id].present? ? BankAccount.find(params[:id]) : current_user.build_bank_account
```

app/observers/project_observer.rb
```diff
@@ -96,7 +96,13 @@ def notify_users(project)
 
   def request_refund_for_failed_project(project)
     project.payments.with_state('paid').each do |payment|
-      payment.request_refund
+      if payment.slip_payment?
```
diogob: this logic should be in Payment model.
we can code it as a condition in the request_refund transition:
```ruby
class Payment < ActiveRecord::Base
  ...
  state_machine :state, initial: :pending do
    ...
    event :request_refund do
      transition paid: :pending_refund, if: ->(payment) {
        payment.can_request_refund?
      }

    end
    ...
  end
  ...
  private
  def can_request_refund?
    !payment.slip_payment? || payment.user.try(:bank_account).try(:valid?)
  end
end
```

app/controllers/admin/contributions_controller.rb
```diff
@@ -5,7 +5,7 @@ class Admin::ContributionsController < Admin::BaseController
   before_filter :set_title
 
   def self.contribution_actions
-    %w[pay refuse refund trash request_refund].each do |action|
```
diogob: request_refund needs to stay :\

diogob: Let's also add chargeback

app/controllers/bank_accounts_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    user.decorator
+  end
+
+  protected
+
+  def user
+    @user ||= resource.user
+  end
+
+  def need_pending_refunds
```
adrianob: This will crash if the user is not logged in.

diogob: Let's add a context "when user is not logged in" in this controller's specs.

app/views/catarse_bootstrap/bank_accounts/edit.html.slim
```diff
@@ -0,0 +1,50 @@
+- content_for :title, t('.page-title')
+
+= render 'shared/header'
+
+.w-section.section-one-column.u-margintop-40
+  .w-container
+    .w-row.u-text-center.u-marginbottom-40
+      .w-col.w-col-2
```
diogob: should be replaced by .w-col-offset-2

app/views/catarse_bootstrap/bank_accounts/edit.html.slim
```diff
@@ -0,0 +1,50 @@
...
+            .w-col.w-col-6
+              = bform.label :owner_name, class: 'field-label fontweight-semibold force-text-dark'
+              = bform.input_field :owner_name, as: :string
+          .w-row
+            .w-col.w-col-6.w-sub-col
+              = bform.label :owner_document, class: 'field-label fontweight-semibold force-text-dark'
+              = bform.input_field :owner_document, as: :tel, validation_text: true, data: { :'validate-cpf-cnpj' => true }
+            .w-col.w-col-6
+          .w-row
+            .w-col.w-col-4
```
diogob: should be replaced by .w-col-offset-4

app/views/catarse_bootstrap/bank_accounts/show.html.slim
```diff
@@ -0,0 +1,28 @@
+- content_for :title, t('.page-title')
+
+= render 'shared/header'
+.w-section.section-one-column.u-margintop-30
+  .w-container
+    .u-text-center.u-marginbottom-10
+      span.fa.fa-check-circle.text-success.fa-3x
+    .w-row.u-text-center
+      .w-col.w-col-1
```
diogob: Let's use offsets here

app/views/catarse_bootstrap/bank_accounts/show.html.slim
```diff
@@ -0,0 +1,28 @@
...
+      .w-col.w-col-10
+        .fontsize-larger.u-marginbottom-20= t('.success')
+      .w-col.w-col-1
+    .w-row.u-text-center.u-marginbottom-40
+      .w-col.w-col-2
```
diogob: Let's use offsets here

app/views/catarse_bootstrap/bank_accounts/show.html.slim
```diff
@@ -0,0 +1,28 @@
...
+      .w-col.w-col-8
+        .fontsize-base= t('.success_detail')
+      .w-col.w-col-2
+    .w-row
+      .w-col.w-col-3
```
diogob: Let's use offsets here

app/controllers/bank_accounts_controller.rb
devton: done ;)

app/controllers/admin/contributions_controller.rb
devton: done ;)

PR #1538, opened 2015-06-22
Revert "If contributor name is nil we should have a default value for…

carmenmp:
Reverts catarse/catarse#1537


diogob:
Hi there, thanks for the PR. But would you care to explain the reasoning for reverting this commit ?

PR #1543, opened 2015-06-23
Refunded confirmation emails

lerrua:


app/views/catarse_bootstrap/user_notifier/mailer/refund_completed_slip.html.slim
```diff
@@ -8,11 +8,9 @@
 |Olá, <strong>#{contribution.user.display_name}</strong>!
 br
 br
-|O reembolso do seu apoio para o projeto #{link_to(contribution.project.name, project_by_slug_url(permalink: contribution.project.permalink))} no #{company_name} acabou de ser realizado para a conta bancária:
+|A devolução do valor do seu apoio para o projeto #{contribution.project.name} no Catarse acabou de ser realizada para a conta bancária:
```
devton: use CatarseSettings[:company_name] instead Catarse name
and use project_by_slug_url when is reference to a project

PR #1558, opened 2015-06-26
remove cursed spec

adrianob:


diogob:
The power of christ compels you!

demon forever-908964

PR #1562, opened 2015-06-26
Lateral bar

adrianob:


app/views/catarse_bootstrap/projects/_dashboard_nav.html.slim
```diff
@@ -48,5 +50,6 @@ nav.w-section.dashboard-nav.side
           = t('.publish')
           | &nbsp;&nbsp;
           span.fa.fa-chevron-right
-
-
+- if policy(@project).update? && @project.published?
+  div
+    a.btn-dashboard.open.fa.fa-lg.fa-chevron-left
```
vicnicius: Here we should use a button element instead of a link. 

`btn.btn-dashboard.open.fa.fa-lg.fa-chevron-left`

PR #1564, opened 2015-06-27
Weekly emails for pending refunds

lerrua:


spec/models/contribution_spec.rb
```diff
@@ -198,4 +198,47 @@
       it{ is_expected.to eq false }
     end
   end
+
+  describe "#need_notify_about_pending_refund" do
```
diogob: should be 
```ruby
describe ".need_notify_about_pending_refund"
```

PR #1573, opened 2015-07-01
Refund flow hotfixes

devton:


config/initializers/sendgrid.rb
```diff
@@ -1,5 +1,5 @@
 begin
-  if Rails.env.production?
+  if Rails.env.production? || Rails.env.sandbox?
```
diogob: wouldn't it be easier to just configure the sandbox RAILS_ENV as production ?

config/initializers/sendgrid.rb
```diff
@@ -14,6 +14,6 @@
   nil
 end
 
-if Rails.env.sandbox?
-  ActionMailer::Base.register_interceptor(SandboxMailInterceptor)
-end
+#if Rails.env.sandbox?
+#  ActionMailer::Base.register_interceptor(SandboxMailInterceptor)
```
diogob: If we are not going to use we should remove the code for SandboxMailInterceptor

PR #1581, opened 2015-07-01
Update README.md

diff-:
To bundle, it needs to gem

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ The official repo is https://github.com/catarse/catarse
 To run this project you need to have:
 
 * Ruby 2.1.2
+* `ruby-devel`
 * [PostgreSQL](http://www.postgresql.org/)
   * OSX - [Postgres.app](http://postgresapp.com/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,7 @@ The official repo is https://github.com/catarse/catarse
 To run this project you need to have:
 
 * Ruby 2.1.2
+* `ruby-devel`
```
diogob: Actually the code runs on older versions as long as you adjust the Gemfile accordingly. I believe that this should be explicit in the README as well.

README.md
```diff
@@ -46,7 +47,9 @@ To run this project you need to have:
 
 * Install the gems
 
-        $ bundle install
+        $ gem install bundler
+        $ gem install json
+        $ bundle install --deployment --without development test
```
diogob: These instructions should be generic and not specific for production use.
You can add something like:
```
bundle install (if you are installing a production environment it's advisable to use --deployment --without development test
```

PR #1591, opened 2015-07-02
Refund bank list

devton:


spec/models/bank_account_spec.rb
```diff
@@ -34,4 +36,46 @@
...
+  end
+
+  describe "#load_bank_from_input_bank_number" do
+    before do
+      custom_bank
+    end
+
+    context "should be called when use .valid?" do
+      before do
+        expect(bank_account).to receive(:load_bank_from_input_bank_number)
```
diogob: this line does can be in the before hook of the parent block

spec/models/bank_account_spec.rb
```diff
@@ -34,4 +36,46 @@
...
+      it { is_expected.to be_truthy }
+    end
+  end
+
+  describe "#load_bank_from_input_bank_number" do
+    before do
+      custom_bank
+    end
+
+    context "should be called when use .valid?" do
```
diogob: We do not need a  context for this. Let's move this spec to a ```describe "#valid?"```

spec/models/bank_account_spec.rb
```diff
@@ -34,4 +36,46 @@
...
+      before do
+        expect(bank_account).to receive(:load_bank_from_input_bank_number)
+      end
+
+      it { bank_account.valid?}
+    end
+
+    it do
```
diogob: We need a description here, could be: "should assign bank according to bank number"

PR #1601, opened 2015-07-07
We should not move payments to deleted anymore

devton:
- removed task that moves payment to trash
- removed psql function can_delete


diogob:
LGTM!
[LGTM](http://lgtm.in/i/697tASGbq)

[:+1:](http://lgtm.in/u/697tASGbq) [:-1:](http://lgtm.in/r/697tASGbq)

PR #1602, opened 2015-07-07
Force element hidden

vicnicius:


app/assets/stylesheets/catarse_bootstrap/_extra.sass
```diff
@@ -405,3 +405,6 @@ img, iframe
   100% 
     +transform(translateY(-50%) scale(0))
     opacity: 0
+    
+.w-hidden
+  display: none !important
```
diogob: Are you sure this won't break any JS that hides and shows elements dynamically ?

vicnicius: great catch, @diogob. we shouldn't display elements based on JS but we sure do.

vicnicius:
Just making sure w-hidden CSS class prevails over others.

PR #1605, opened 2015-07-08
Fix newsletter subscribe submit

lerrua:


app/assets/javascripts/app/pages/home.js
```diff
@@ -5,4 +5,18 @@ App.addChild('Home', {
     $.animateHeadline();
   },
 
+});
+
+App.addChild('HomeNewsletter', {
```
diogob: Why not use a input type submit instead of the JS ?

PR #1623, opened 2015-07-14
remove duplicated cron job

adrianob:


lib/tasks/cron.rake
```diff
@@ -4,8 +4,7 @@ namespace :cron do
                 :refresh_materialized_views]
 
   desc "Tasks that should run daily"
-  task daily: [:refresh_materialized_views,
-               :notify_project_owner_about_new_confirmed_contributions,
+  task daily: [ :notify_project_owner_about_new_confirmed_contributions,
```
diogob: daily should depend on refresh_materialized_views

PR #1631, opened 2015-07-15
New posts

adrianob:


app/views/catarse_bootstrap/projects/_dashboard_posts.html.slim
```diff
@@ -1,16 +1,16 @@
-.section
+.w-section.section
   .w-container
     .w-row
-      .w-col.w-col-8
-        .project-news.u-marginbottom-60
-          .fontsize-large.fontweight-semibold.u-marginbottom-30= t('.title')
-          = simple_form_for @project,  html: { class: 'w-form card card-terciary' } do |f|
+      .w-col.w-col-1
```
vicnicius: I think we should remove this empty w-col-1 div and add .w-col-push-1 to the colune where the content is.

app/views/catarse_bootstrap/projects/_dashboard_posts.html.slim
```diff
@@ -19,14 +19,15 @@
 
               .card.card-message.u-radius.u-marginbottom-40
                 .fontsize-small= t('.warning')
-              .w-row
-                .w-col.w-col-4.w-sub-col.u-marginbottom-20
+
+              .w-row.u-marginbottom-20
+                .w-col.w-col-2
+                .w-col.w-col-5.w-sub-col
```
vicnicius: Again, here we should use w-col-push-2 to the w-col-5 and remove the empty w-col-2 div.

PR #1633, opened 2015-07-15
Syncs project nav-bar animation with body CSS class transition

vicnicius:


diogob:
Makes more sense :D

vicnicius:
[fix #98376312] [fix #98293982] [fix #98293602]

PR #1663, opened 2015-07-31
Catarse members

devton:


diogob:
Let's try to unify the common JS requirements and avoid the current code duplication on init.js and application.js

db/migrate/20150729110132_add_team_members_view.rb
```diff
@@ -0,0 +1,97 @@
+class AddTeamMembersView < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+      CREATE ROLE anonymous NOLOGIN;
+      CREATE OR REPLACE FUNCTION was_confirmed(contributions) RETURNS boolean
```
diogob: should be public. was_confirmed

db/migrate/20150729110132_add_team_members_view.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def down
+    execute <<-SQL
+      DROP ROLE anonymous NOLOGIN;
+      drop view "1".team_members;
+      drop view if exists "1".team_totals;
+      drop index if exists user_totals_user_id_ix;
+      drop index if exists user_admin_id_ix;
+
+      CREATE OR REPLACE FUNCTION was_confirmed(contributions) RETURNS boolean
```
diogob: should be public.was_confirmed

PR #1669, opened 2015-08-04
do not reset user chosen value when clicking on another reward

adrianob:


app/assets/javascripts/app/contribution/contribution.js
```diff
@@ -16,7 +16,11 @@ App.addChild('Contribution', {
   submitForm: function(event){
     var $target_row = $(event.target).parents('.back-reward-money'),
         user_value = this.$('.selected').find('.user-reward-value').val().replace(/\./g,'');
-    this.$value.val(user_value);
+    if(user_value === ''){
```
diogob: I'd say we also need to reser when ```this.$value.val() < this.minimumValue()```

adrianob: If we did this there would be no error when the user chooses a smaller value. The user would proceed to the next screen with a different value than he chose, and he might not notice the change.

PR #1686, opened 2015-08-11
add anonymous toggle to user dashboard

adrianob:


app/controllers/projects/contributions_controller.rb
```diff
@@ -71,6 +71,12 @@ def second_slip
     redirect_to resource.details.ordered.first.second_slip_path
   end
 
+  def toggle_anonymous
+    authorize resource
+    contribution = Contribution.find params[:id]
```
devton: @adrianob i think that you don't need of this contribution just use resource

PR #1693, opened 2015-08-13
FIX: SQL script with IF NOT EXIST CREATE ROLE

dalink:


diogob:
Thanks for the PR :+1:

PR #1697, opened 2015-08-14
FIX french translation issue

dalink:


config/locales/catarse_bootstrap/views/projects/edit.pt.yml
```diff
@@ -175,8 +175,6 @@ pt:
     dashboard_card:
       submit: Salvar
       card_label: "Card do projeto"
-    dashboard_goal:
```
adrianob: Why remove this?

dalink: Because it's a duplicate line (see line 155)

dalink:
Thanks

PR #1698, opened 2015-08-17
send second slip notification on cron task

adrianob:


lib/tasks/cron.rake
```diff
@@ -15,6 +15,16 @@ namespace :cron do
     UserTotal.refresh_view
   end
 
+  desc "Send second slip notification"
+  task second_slip_notification: :environment do
+    puts "sending second slip notification"
+    ContributionDetail.where(payment_method: 'BoletoBancario',
```
diogob: This ```where``` should be a scope in the model with an accompanying spec. Something like ```slips_past_waiting``` or ```slips_waiting_forever```

app/models/contribution_detail.rb
```diff
@@ -39,6 +39,13 @@ class ContributionDetail < ActiveRecord::Base
     where("contribution_details.state not in('deleted', 'refused', 'pending') OR payments.waiting_payment")
   }
 
+  scope :slips_past_waiting, -> {
+    where(payment_method: 'BoletoBancario',
+          state: 'pending',
+          waiting_payment: false,
```
devton: nice new thing learned today <3

PR #1710, opened 2015-08-20
Bugfixes

devton:
- check if payment already paid before sent confirmation email
- show pending slip when project is online at user dash

### DIFF
--- a/app/models/contribution_detail.rb
+++ b/app/models/contribution_detail.rb
@@ -31,7 +31,19 @@ class ContributionDetail < ActiveRecord::Base
   # Scopes based on project state
   scope :with_project_state, ->(state){ where(project_state: state) }
   scope :for_successful_projects, -> { with_project_state('successful').available_to_display }
-  scope :for_online_projects, -> { with_project_state(['online', 'waiting_funds']).available_to_display }
+  scope :for_online_projects, -> {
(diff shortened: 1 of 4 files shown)


app/models/contribution_detail.rb
```diff
@@ -31,7 +31,19 @@ class ContributionDetail < ActiveRecord::Base
...
   scope :with_project_state, ->(state){ where(project_state: state) }
   scope :for_successful_projects, -> { with_project_state('successful').available_to_display }
-  scope :for_online_projects, -> { with_project_state(['online', 'waiting_funds']).available_to_display }
+  scope :for_online_projects, -> {
+    with_project_state(['online', 'waiting_funds']).
+    where(
+    <<-SQL
+      case
+      when contribution_details.payment_method = 'CartaoDeCredito' then
+        contribution_details.state not in('deleted', 'refused', 'pending')
```
diogob: Why make this distinction

app/models/contribution_detail.rb
```diff
@@ -31,7 +31,19 @@ class ContributionDetail < ActiveRecord::Base
...
+      else
+        contribution_details.state not in('deleted')
```
diogob: It seems that we should only ignore deleted in both cases

PR #1713, opened 2015-08-21
New reports metrics [Epic ##2012754]

devton:
- added chartjs require dependency
- added project-insights component
- reports tab show only reports to download
- added project_details view for API handles
- added project_contributions_per_location view for API handles
- added project_contributions_per_day view for API handles


db/migrate/20150819150242_add_more_fields_to_project_details.rb
```diff
@@ -0,0 +1,133 @@
...
+        LANGUAGE sql STABLE SECURITY DEFINER
+          AS $$
+            SELECT '{"online", "waiting_funds", "failed", "successful"}'::text[];
+          $$;
+
+      drop function if exists public.is_published(projects);
+      create function public.is_published(projects) returns boolean
+        language sql stable SECURITY DEFINER
+        as $$
+          select true where $1.state = ANY(public.published_states());
```
diogob: This will return NULL for a non published project, it's better to use:
```
select $1.state = ANY(public.published_states());
```

db/migrate/20150819150242_add_more_fields_to_project_details.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      drop function if exists public.remaining_time_interval(projects);
+      CREATE FUNCTION public.remaining_time_interval(projects) RETURNS interval
+        LANGUAGE sql STABLE SECURITY DEFINER
+          AS $$
+            select ($1.expires_at - current_timestamp)::interval
+          $$;
+
+      drop function if exists public.remaining_time_json(projects);
+      CREATE FUNCTION public.remaining_time_json(projects) RETURNS json
```
diogob: Nice one :beers:

db/migrate/20150819150242_add_more_fields_to_project_details.rb
```diff
@@ -0,0 +1,133 @@
...
+    execute <<-SQL
+      drop function if exists public.published_states();
+      CREATE FUNCTION public.published_states() RETURNS text[]
+        LANGUAGE sql STABLE SECURITY DEFINER
+          AS $$
+            SELECT '{"online", "waiting_funds", "failed", "successful"}'::text[];
+          $$;
+
+      drop function if exists public.is_published(projects);
+      create function public.is_published(projects) returns boolean
```
diogob: We should rewrite the model method:
``` 
published?
```
using this function to avoid duplicated logic between database and ruby

db/migrate/20150819150242_add_more_fields_to_project_details.rb
```diff
@@ -0,0 +1,133 @@
...
+        language sql stable SECURITY DEFINER
+        as $$
+          select true where $1.state = ANY(public.published_states());
+        $$;
+
+      drop function if exists public.is_expired(projects);
+      CREATE FUNCTION public.is_expired(projects) RETURNS boolean
```
diogob: We have a scope in the ruby model called ```expired```. We should rewrite that scope to use this function and avoid duplicate logic (besides, we can reuse the same specs in this way).

db/migrate/20150819150242_add_more_fields_to_project_details.rb
```diff
@@ -0,0 +1,133 @@
...
+        LANGUAGE sql STABLE SECURITY DEFINER
+          AS $$
+            SELECT (current_timestamp > $1.expires_at);
+          $$;
+
+      drop function if exists public.open_for_contributions(projects);
+      CREATE FUNCTION public.open_for_contributions(projects) RETURNS boolean
```
diogob: We have ```accept_contributions?``` in the model, should rewrite using this.

db/migrate/20150819150242_add_more_fields_to_project_details.rb
diogob: Let's try to unify this function and the one we have in the ```ProjectDecorator``` called ```time_to_go```. It's better to have a standard for this string and also not duplicate logic

db/migrate/20150819150242_add_more_fields_to_project_details.rb
lerrua: :+1:

PR #1714, opened 2015-08-25
adjusting devise

thiagocatarse:


thiagocatarse:
I did some mistakes while I was creating the pull request, so I guess one of the commits will be a little weird (it looks like I deleted everything and rewrote again).. But I imagine that everything should be fine with this PR...

If someone can review it would be great. We are basically just moving the question (Already registered or Don't have an account) to close to the submit button.

PR #1730, opened 2015-08-31
Mixpanel insights

lerrua:


vicnicius:
@lerrua Concerning to this PR, I was exploring Mixpanel and it looks like we are not tracking this event correctly yet. We'll have to debug it, so while we don't do this, I'll add a dont_merge tag to it, ok?

lerrua:
okay @vicnicius

PR #1742, opened 2015-09-08
Mithril home

adrianob:


db/migrate/20150902210721_create_projects_view.rb
```diff
@@ -0,0 +1,51 @@
+class CreateProjectsView < ActiveRecord::Migration
+  def up
+     execute <<-SQL
+
+       CREATE OR REPLACE FUNCTION public.img_thumbnail(projects, size text)
+       RETURNS text
+       LANGUAGE sql
+       STABLE
```
devton: I think that you need to use security definer here.

diogob: We do not, settings have security definer already, so this functions does not access the database by itself :D

db/migrate/20150908182326_update_projects_for_home.rb
```diff
@@ -0,0 +1,150 @@
...
+              recommends.expires_at,
+              img_thumbnail(recommends.*,'large') AS project_img,
+
+              pt.pledged,
+              pt.progress,
+              s.acronym as state_acronym,
+              u.name AS owner_name,
+              c.name AS city_name
+
+             FROM (projects recommends
```
diogob: We need to qualify all table references from now on. This should be ```public.projects```

db/migrate/20150908182326_update_projects_for_home.rb
```diff
@@ -0,0 +1,150 @@
...
+              JOIN users u ON (recommends.user_id = u.id)
```
diogob: ```public.users```

db/migrate/20150908182326_update_projects_for_home.rb
```diff
@@ -0,0 +1,150 @@
...
+              LEFT JOIN "1".project_totals pt ON (pt.project_id = recommends.id)
+              LEFT JOIN cities c ON (c.id = recommends.city_id)
```
diogob: ```public.cities``` and ```public.states```

app/views/catarse_bootstrap/projects/index.html.slim
```diff
@@ -19,48 +19,50 @@ section#hero-home.hero-full.hero-knowmore
...
-      .w-container
-        .w-row.u-marginbottom-30
-          .w-col.w-col-10.w-col-small-6.w-col-tiny-6
-            .fontsize-large.lineheight-looser
-              = t('.near_you')
-          .w-col.w-col-2.w-col-small-6.w-col-tiny-6
-            = link_to t('.see_all'), explore_path(ref: 'home_near', anchor: "near_of/" ) + current_user.address_state, class: 'btn btn-small btn-terciary'
-        .w-row
-          == render_projects @projects_near, 'home_near', { wrapper_class: 'w-col w-col-4' }
+#project-index-root
```
diogob: Let's remove this commented code

app/models/projects_for_home.rb
```diff
@@ -1,12 +0,0 @@
-class ProjectsForHome < Project
```
diogob: Nice! :+1:

PR #1757, opened 2015-09-21
fix limit reward checkbox [#103256262]

adrianob:


diogob:
rspec ./spec/decorators/project_decorator_spec.rb:95 # ProjectDecorator#display_expires_at when we have an online_date should eq "22/09/2015"

This might have something to do with the use of the database timestamp.

PR #1759, opened 2015-09-21
fix project image on quick search [#102825512]

adrianob:


app/assets/stylesheets/catarse_bootstrap/_extra.sass
```diff
@@ -13,6 +13,8 @@ img, iframe
 .list-answer
   display: none
 
+.project-search-image
+  max-width: 50px !important
```
diogob: Why use important here?

adrianob: Otherwise it will be overwritten by the `max-width: 100%` on `img` right above.

diogob: I see, but the fix should remove the !important from img so the CSS would prefer the max-width from the class.
!important is a terrible contagious hack, the more we use it, the more we will have to use it.

diogob: the !important in the img was added by @vicnicius here: 265eba603d89b6ccae6fa1d6c5384a711466e692

Maybe this rule could be more specific. I believe this has to do with redactor and the use of imager embeds.

adrianob: @diogob I couldn't figure out why it was there in the first place, everything seems fine with redactor without using !important. I have removed it for now, we can review this later if any problems arise.

PR #1778, opened 2015-09-29
Mithril explore

adrianob:


db/migrate/20150924220114_create_categories_view.rb
```diff
@@ -0,0 +1,33 @@
...
+           name_pt,
+           name_en,
+           (SELECT count(*) from projects p where p.state = 'online' AND p.category_id = c.id) as online_projects,
+           (SELECT count(DISTINCT user_id) from category_followers cf where cf.category_id = c.id) as followers
+
+         FROM categories c
+         WHERE
+          exists(select true from projects p where p.category_id = c.id and p.state not in('draft', 'rejected'))
+          GROUP BY c.id
+          ORDER BY name_pt;
```
diogob: let's drop the order by, we can ask for the order using the API

db/migrate/20150924220114_create_categories_view.rb
```diff
@@ -0,0 +1,33 @@
...
+        SELECT c.id,
+           name_pt,
+           name_en,
+           (SELECT count(*) from projects p where p.state = 'online' AND p.category_id = c.id) as online_projects,
+           (SELECT count(DISTINCT user_id) from category_followers cf where cf.category_id = c.id) as followers
+
+         FROM categories c
+         WHERE
+          exists(select true from projects p where p.category_id = c.id and p.state not in('draft', 'rejected'))
+          GROUP BY c.id
```
diogob: we can drop this group by for it's redundant (the counts are subqueries anyway)

db/migrate/20150924220114_create_categories_view.rb
```diff
@@ -0,0 +1,33 @@
+class CreateCategoriesView < ActiveRecord::Migration
+  def change
+    create_table :categories_views do |t|
+      execute "
+      CREATE OR REPLACE VIEW \"1\".categories AS
+        SELECT c.id,
+           name_pt,
```
diogob: man, let's use just the ```name_pt as name```

diogob:
moved to mithril-explore

PR #1781, opened 2015-10-01
Increase minimum postgresql version to 9.4

ArnisL:
Stuff like json_build_object works starting from postgres version 9.4

Comment about postgresql version in sample database.yml
is misleading.


diogob:
Thanks for the PR :+1:

PR #1788, opened 2015-10-02
Donation screen

adrianob:


app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,27 @@
+class DonationsController < ApplicationController
+
+  helper_method :resource
+
+  def create
+    @donation = Donation.create
+    raise Pundit::NotAuthorizedError if !current_user
```
diogob: this should be the first line

app/observers/payment_observer.rb
```diff
@@ -18,6 +18,7 @@ def from_chargeback_to_paid(payment)
   end
 
   def from_pending_refund_to_refunded(payment)
+    return if Contribution.where.not(donation:nil).where(id: payment.contribution).exists?
```
diogob:
```ruby
return if payment.contribution.donation.present?
```

devton: maybe we can create a method on contribution like 
```ruby
def is_donation?
   self.donation.exists?
end
```
and delegate to payment :)

diogob: better yet!

adrianob: I have just realised all donations will go from paid to refunded directly, so we don't need this check at all :)

adrianob: Actually we do need it, just saw the alias below.

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    @donation
+  end
+
+  private
+  def update_pending_refunds
+    current_user.pending_refund_payments.each do |payment|
+      payment.contribution.update_attribute :donation, @donation
+
+      payment.state = 'refunded'
+      payment.save!
```
devton: @diogob @vicnicius @adrianob we need to create a badge on contributions admin to show to support team that this these contributions is donations, because when they look at payment gateway dashboard the state is 'paid'

db/migrate/20151002164325_create_donation_notifications.rb
```diff
@@ -0,0 +1,14 @@
+class CreateDonationNotifications < ActiveRecord::Migration
+  def change
+    create_table :donation_notifications do |t|
+      t.integer :user_id, null: false
+      t.integer :donation_id, null: false
+      t.text :from_email, null: false
+      t.text :from_name, null: false
+      t.text :template_name, null: false
+      t.text :locale, null: false
+      t.timestamp :sent_at
```
devton: missing column deliver_at check out another *_notifications table

adrianob: done

PR #1798, opened 2015-10-07
PostgreSQL version

hallison:
Catarse requires PostgreSQL v9.4 which includes improvements in window functions and filtering for aggregation.


devton:
<3

PR #1800, opened 2015-10-07
Update donations

adrianob:
Implements donations and bank account transfers.
Transfers are used for legacy payments, and donations can come from both pagarme and legacy contributions.


app/models/user.rb
```diff
@@ -153,9 +153,7 @@ def pending_refund_payments
       projects: {
         state: 'failed'
       },
-      state: 'paid',
-      gateway: 'Pagarme',
```
adrianob: Why we had this in the first place?

diogob: @adrianob @devton should know better, I believe he created this scope

diogob: I believe there are other parts of the system that expect this payment to be done through pagarme. The best approach here is to split this scope in 2.

adrianob: This was done since we can't refund payments from other gateways. I will keep this scope generic and update our request_refund method to use a bank account transfer instead of a refund for legacy gateways.

db/migrate/20151008235831_update_user_totals.rb
```diff
@@ -0,0 +1,113 @@
...
+          sum(
+              CASE
+          WHEN lower(pa.gateway) = 'pagarme'::text THEN 0::numeric
+          WHEN p.state::text <> 'failed'::text AND NOT uses_credits(pa.*) THEN 0::numeric
+          WHEN p.state::text = 'failed'::text AND uses_credits(pa.*) THEN 0::numeric
+          WHEN p.state::text = 'failed'::text AND ((pa.state = ANY (ARRAY['pending_refund'::character varying::text, 'refunded'::character varying::text])) AND NOT uses_credits(pa.*) OR uses_credits(pa.*) AND NOT (pa.state = ANY (ARRAY['pending_refund'::character varying::text, 'refunded'::character varying::text]))) THEN 0::numeric
+          WHEN p.state::text = 'failed'::text AND NOT uses_credits(pa.*) AND pa.state = 'paid'::text THEN pa.value
+          ELSE pa.value * (-1)::numeric
+              END
+          ) - (SELECT sum(amount)/100 from user_transfers ut WHERE ut.status = 'transferred' AND ut.user_id = c.user_id) AS credits
```
adrianob: Here is the relevant change @diogob @devton

diogob:
moved to catarse/update_donations

PR #1805, opened 2015-10-08
Adds project-image css class to extra

vicnicius:
@thiagocatarse I'm adding this class to the _extra while we don't update our webflow.css. As soon as we do it, we should remove it from extras.


thiagocatarse:
ok @vicnicius !

PR #1827, opened 2015-10-16
Add deactivated_at column into user_details

lerrua:


db/migrate/20151016212000_add_deactivated_at_to_user_details.rb
```diff
@@ -0,0 +1,64 @@
...
+    execute <<-SQL
+      DROP VIEW "1".user_details;
+
+      CREATE VIEW "1".user_details AS
+        SELECT u.id,
+          u.name,
+          u.address_city,
+          u.deactivated_at,
+          thumbnail_image(u.*) AS profile_img_thumbnail,
+          u.facebook_link,
```
diogob: facebook link should be for owner or admin

PR #1831, opened 2015-10-19
Flexible projects basics

devton:
- Added flexible project (model) to handle with projects that are flexible
- Drop video_url validation
- Remove reward size validation when flexible project
- Flexible project should not fail


app/models/flexible_project.rb
```diff
@@ -0,0 +1,7 @@
+class FlexibleProject < ActiveRecord::Base
```
diogob: let's move this to a type column.

app/models/project.rb
```diff
@@ -197,7 +199,7 @@ def new_draft_recipient
   end
 
   def should_fail?
```
diogob: Let's move this to a separate object, like the state machine one

app/models/concerns/project/state_machine_handler.rb
```diff
@@ -21,17 +22,11 @@ module Project::StateMachineHandler
             self.user.errors.add_on_blank(attr)
           end
           self.user.errors.each {|error, error_message| self.errors.add('user.' + error.to_s, error_message)}
-          self.errors['rewards.size'] << "Deve haver pelo menos uma recompensa" if self.rewards.size == 0
+          self.errors['rewards.size'] << "Deve haver pelo menos uma recompensa" if self.rewards.size == 0 && !self.is_flexible?
```
diogob: Let's use 2 state machines, one deriving from the other

app/models/concerns/project/state_machine_handler.rb
```diff
@@ -82,7 +77,7 @@ module Project::StateMachineHandler
         }
 
         transition waiting_funds: :successful,  if: ->(project) {
-          project.reached_goal?
+          project.reached_goal? || project.is_flexible?
```
diogob: Let's use 2 state machines, one deriving from the other

diogob:
Moving to catarse branch

PR #1844, opened 2015-10-23
add donation confimation screen

adrianob:
screen shot 2015-10-23 at 4 20 16 pm

### DIFF
--- a/app/views/catarse_bootstrap/donations/confirm.html.slim
+++ b/app/views/catarse_bootstrap/donations/confirm.html.slim
@@ -1,0 +1,22 @@
+= render 'shared/header'
+
+#contribution-donation
+  .w-section.section-one-column.u-margintop-40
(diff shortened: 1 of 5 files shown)


app/views/catarse_bootstrap/donations/confirm.html.slim
```diff
@@ -0,0 +1,22 @@
+= render 'shared/header'
+
+#contribution-donation
+  .w-section.section-one-column.u-margintop-40
+    .w-container
+      .w-row.u-text-center.u-marginbottom-60
+        .w-col.w-col-2
+        .w-col.w-col-8
+          .fontsize-larger.u-marginbottom-20 Confirma sua doação para o #{CatarseSettings[:company_name]}?
```
devton: lets keep these texts on I18n :)

PR #1847, opened 2015-10-26
fix user transfer rake task

adrianob:


lib/tasks/verify_pagarme_transactions.rake
```diff
@@ -26,7 +26,7 @@ task verify_pagarme_user_transfers: [:environment] do
 
     payment_transfer.update_attribute(:transfer_data, transfer.to_hash)
     if transfer.status == 'failed'
-      transfer.user.contributions.last.invalid_refund
+      payment_transfer.user.contributions.last.invalid_refund
```
diogob: I think we should attach the notification to the UserTransfer or add the contribution id to it.
In the way that is coded we loose important information.

adrianob: It is not related to a specific contirbution. I've added a new notifications table so we can keep track of what we send.

PR #1852, opened 2015-10-29
Migrates delete_project_reminder from PL/pgSQL to PL/v8

diogob:


diogob:
So apparently PL/v8 functions are not clear winners. SQL execution looks clumsier, it also lacks support for ES6 in the current version.
Besides a short benchmark I did showed that PL/pgSQL will perform much better in some cases (although the same can be said about PL/v8).
@lerrua @adrianob @devton and I have decided to keep functions in PL/pgSQL for now.

devton:
SQL FTW

PR #1867, opened 2015-11-06
Adds project detail contribution per referral view

vicnicius:
"WTF?! The frontend guy is writing SQL?!"

No, I'm not! Just needed to do some tests in a component and ended up writing this. As it was working locally, thought that maybe it could save you guys some time. Copied the ideas from similar views, of course x).


diogob:
Beautiful! Now to win our hearts completely you can deliver a neat spec for this endpoint in this repo: https://github.com/catarse/catarse-api-specs

PR #1870, opened 2015-11-07
Fixes project goal mask

vicnicius:
[fix #105243462] [fix #106483654]


diogob:
This mask needs to be correctly interpreted in the backend, hence the use of ```include I18n::Alchemy ``` in the Contribution model. I believe we need to add the same feature for the project model.

vicnicius:
@diogob done.

PR #1872, opened 2015-11-09
Adds is_flexible? and Statesmen gem 

diogob:
Breaking giant mothefucka PR in smaller chunks!
dengeki sentai changeman ep 01 shutsugen himitsu no chikara mkv_snapshot_17 04_ 2013 08 25_10 36 52

### DIFF
--- a/app/models/concerns/project/state_validator.rb
+++ b/app/models/concerns/project/state_validator.rb
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# This module handles with default state validation
+module Project::StateValidator
+  extend ActiveSupport::Concern
+
+  included do
+
+    # All valid states for projects in_analysis to end of publication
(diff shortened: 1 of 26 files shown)


app/models/concerns/project/state_validator.rb
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# This module handles with default state validation
+module Project::StateValidator
+  extend ActiveSupport::Concern
+
+  included do
+
+    # All valid states for projects in_analysis to end of publication
+    ON_ANALYSIS_TO_END_STATES = %w(in_analysis approved online successful waiting_funds failed).freeze
```
adrianob: Can't we get this from the db now?

diogob: Yeah, I just kept this code here, because I think it's easier to first merge and then we refactor

PR #1883, opened 2015-11-13
Flexible in dashboard

devton:


app/views/catarse_bootstrap/projects/_dashboard_announce_expiration.html.slim
```diff
@@ -0,0 +1,20 @@
+.w-row.u-marginbottom-60
+  .w-col.w-col-1
+  .w-col.w-col-10
+    .card.card-big.card-terciary.u-radius
+      .fontsize-base= t('.explanation_01')
+      .u-text-center.fontsize-large.u-marginbottom-40.u-margintop-40
+        span.card.u-radius
+          = l(7.days.from_now.to_date)
```
diogob: what does this hardcoded 7 means? If it should be hardcoded is better to use a constant

app/views/catarse_bootstrap/projects/_dashboard_goal.html.slim
```diff
@@ -18,18 +18,27 @@
...
-                append_input_text: 'Dias',
-                append_input_text_wrapper_html: { class: 'fontsize-base lineheight-tightest' }
+              = render "#{@project.mode}_dashboard_explanation"
+
+              - if @project.is_flexible?
+                .w-row.u-marginbottom-30.card.card-terciary.medium
+                  .w-col.w-col-5.w-sub-col._w-sub-col
+                    label.w-editable.fontweight-semibold.fontsize-base= t(".#{@project.mode}.expiration_title")
+                  .w-col.w-col-7
+                    .w-editable.fontsize-small= t(".#{@project.mode}.expiration_text")
```
diogob: This mode logic is duplicated with the above if.
One solution is to use hardcode the I18n keys or create two separate partials.

app/views/catarse_bootstrap/projects/_dashboard_nav.html.slim
```diff
@@ -9,6 +9,11 @@
         - else
           img.thumb-project-dashboard src="#{image_path('thumb-project.png')}" width="114"
         .fontcolor-negative.lineheight-tight.fontsize-small= @project.name
+        - if @project.is_flexible?
```
diogob: I prefer to avoid the if whenever possible and just use the mode.

PR #1884, opened 2015-11-13
save events from sendgrid webhook [#107532360]

adrianob:


devton:
@adrianob we dont need to define an route for receive these events?

adrianob:
@devton No, apparently this gem defines it automatically.

PR #1897, opened 2015-11-17
Updates UIHelper to only resize iframes contained on activation eleme…

vicnicius:
…nt [fix #108287972]


app/assets/javascripts/lib/ui_helper.js
```diff
@@ -1,8 +1,9 @@
 var UIHelper = window.UIHelper = {
   setupResponsiveIframes: function($container) {
     var that = this;
+    console.log('Will implement shit on ', $container);
```
diogob: we don't want shit on our console logs :P

vicnicius: hehehe. shit removed.

PR #1899, opened 2015-11-18
Removes several IFs to mount mithril components and adds generic mounting base on data attributes

diogob:


vicnicius:
Beautiful! :beers:

PR #1907, opened 2015-11-20
Should always update expires_at after save if online_days is defined [fix #108620628]

diogob:


app/models/project.rb
```diff
@@ -284,6 +284,12 @@ def open_for_contributions?
     pluck_from_database(:open_for_contributions)
   end
 
+  def update_expires_at
+    if self.online_days.present?
```
devton: i think that we should check if online_date is present too

PR #1911, opened 2015-11-23
Categories endpoint

diogob:


db/migrate/20151123181718_create_categories_endpoint.rb
```diff
@@ -0,0 +1,18 @@
+class CreateCategoriesEndpoint < ActiveRecord::Migration
+  def change
+    execute <<-SQL
+    CREATE OR REPLACE VIEW "1".categories AS
+    SELECT
+        c.id,
+        name_pt as name,
+        (SELECT count(*) from projects p where p.state = 'online' AND p.category_id = c.id) as online_projects,
```
devton: should look at flexible_projects too :)

diogob: Good catch, I'll use ```open_for_contributions```

PR #1915, opened 2015-11-25
notifications endpoint

lerrua:


diogob:
Let's wait for the JS code before merging this.

lerrua:
The JS code is already done.

PR #1924, opened 2015-12-01
Adds zone_created_at to project_contributions endpoint

vicnicius:


devton:
@diogob @vicnicius what about the timezone on user?

diogob:
@devton I was thinking the same thing. Maybe we should abolish the idea of having the timestamp and the zone_timestamp and just convert all timezones. What do you think guys?

devton:
@diogob yes! :dancer:

vicnicius:
+1 @devton @diogob

diogob:
@vicnicius let's drop the zone_created_at then. just replace the **created_at** with **public.zone_timestamp(c.created_at) as created_at** and it will work for now.

db/migrate/20151201024329_adds_zone_created_at_to_project_contributiosn_endpoint.rb
```diff
@@ -0,0 +1,63 @@
...
+    ut.total_contributed_projects,
+    public.zone_timestamp(c.created_at) AS created_at
+   FROM ((((public.contributions c
+     JOIN public.users u ON ((c.user_id = u.id)))
+     JOIN public.projects p ON ((p.id = c.project_id)))
+     JOIN public.payments pa ON ((pa.contribution_id = c.id)))
+     LEFT JOIN "1".user_totals ut ON ((ut.user_id = u.id)))
+  WHERE ((public.was_confirmed(c.*) OR public.waiting_payment(pa.*)) AND ((NOT c.anonymous) OR public.is_owner_or_admin(p.user_id)));
+
+grant select on "1".project_contributions to admin;
```
diogob: we don't need the grants now that the view interface does not change

db/migrate/20151201024329_adds_zone_created_at_to_project_contributiosn_endpoint.rb
```diff
@@ -0,0 +1,63 @@
+class AddsZoneCreatedAtToProjectContributiosnEndpoint < ActiveRecord::Migration
+ def up
+    execute <<-SQL
+DROP VIEW "1".project_contributions CASCADE;
+CREATE VIEW "1".project_contributions AS
```
diogob: use a create or replace here

db/migrate/20151201024329_adds_zone_created_at_to_project_contributiosn_endpoint.rb
```diff
@@ -0,0 +1,63 @@
+class AddsZoneCreatedAtToProjectContributiosnEndpoint < ActiveRecord::Migration
+ def up
+    execute <<-SQL
+DROP VIEW "1".project_contributions CASCADE;
```
diogob: no need to drop

PR #1930, opened 2015-12-03
New referral structure

devton:
- Added Origins table to store the referral / domain
- Migrate projects and contributions to use origins
- Adjusted API contributions_per_ref to look at origins instead referral_link
- Adjusted DB triggers to use origins

### DIFF
--- a/app/models/origin.rb
+++ b/app/models/origin.rb
@@ -1,4 +1,25 @@
 class Origin < ActiveRecord::Base
   has_many :projects
   has_many :contributions
+
+  validates :domain, presence: true
+  validates :domain, uniqueness: { scope: :referral }
+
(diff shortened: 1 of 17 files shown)


app/models/origin.rb
```diff
@@ -0,0 +1,29 @@
+class Origin < ActiveRecord::Base
+  has_many :projects
+  has_many :contributions
+
+  validates :domain, presence: true
+  validates :domain, uniqueness: { scope: :referral }
+
+  # Process a given ref and http referrer
+  # until found or create a new Origin
+  def self.process referral, http_referrer
```
diogob: Can't we move this logic into self.process_hash?
Or at least make this method private?

PR #1933, opened 2015-12-03
Tag system

adrianob:
Simple comma separated tag system for admin use.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -297,8 +297,8 @@ class Project < ActiveRecord::Base
 
   def all_tags=(names)
     self.tags = names.split(',').map do |name|
-      Tag.find_or_create_by(name: name.strip) do |tag|
-        tag.slug = name.parameterize
+      Tag.find_or_create_by(slug: name.parameterize) do |tag|
+        tag.name = name.strip
       end
     end
   end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -297,7 +297,9 @@ class Project < ActiveRecord::Base
 
   def all_tags=(names)
     self.tags = names.split(',').map do |name|
-      Tag.find_or_create_by(name: name.strip)
+      Tag.find_or_create_by(name: name.strip) do |tag|
+        tag.slug = name.parameterize
+      end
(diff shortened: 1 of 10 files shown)


db/migrate/20151203184721_create_tags.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTags < ActiveRecord::Migration
+  def change
+    create_table :tags do |t|
+      t.text :name
+
+      t.timestamps
```
devton: @adrianob unique index on tag name?  ```add_index :tags, :name, unique: true```

db/migrate/20151203190446_create_taggings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTaggings < ActiveRecord::Migration
+  def change
+    create_table :taggings do |t|
+      t.references :tag
+      t.references :project
+
+      t.timestamps
```
devton: scoped unique index here ```add_index :taggings, [:tag_id, :project_id], unique: true```

app/models/project.rb
```diff
@@ -293,6 +295,16 @@ def open_for_contributions?
     pluck_from_database(:open_for_contributions)
   end
 
+  def all_tags=(names)
+    self.tags = names.split(',').map do |name|
+      Tag.find_or_create_by(name: name.strip)
```
devton: lets force lower case on tag name?

app/models/project.rb
```diff
@@ -293,6 +295,18 @@ def open_for_contributions?
     pluck_from_database(:open_for_contributions)
   end
 
+  def all_tags=(names)
+    self.tags = names.split(',').map do |name|
+      Tag.find_or_create_by(name: name.strip) do |tag|
```
devton: @adrianob we should use ```find_or_create_by(slug: ...)``` here not?

adrianob: The idea is to save the exact user input and a slug to help searching later, if we look for slug first there is no way to recover the information later, that is, there is no way of going from slug to name.

PR #1935, opened 2015-12-04
load zendesk js asynchronously

adrianob:


app/views/catarse_bootstrap/layouts/_new_zendesk.html.slim
```diff
@@ -1,4 +1,4 @@
-script src="//assets.zendesk.com/external/zenbox/v2.6/zenbox.js" type="text/javascript"
+<script async src="//assets.zendesk.com/external/zenbox/v2.6/zenbox.js" type="text/javascript"></script>
```
diogob: Why not use slim here?

adrianob: I could't find a way to set the async attribute with slim. The closest would be setting async=""

PR #1960, opened 2015-12-11
README with postgREST instructions

lerrua:
markdown preview here:
https://github.com/lerrua/catarse/blob/postgrest_readme/README.md


devton:
Lets update the payments gateway info, just pagar.me

PR #1963, opened 2015-12-14
update statistics view

devton:


diogob:
we should add a test case for this on https://github.com/catarse/catarse-api-specs

PR #2003, opened 2016-01-07
add link to admin push to flex

adrianob:


app/controllers/projects_controller.rb
```diff
@@ -100,6 +100,17 @@ def video
     render json: nil
   end
 
+  def push_to_flex
```
devton: gogo powerspecs <3

PR #2013, opened 2016-01-13
Normalize city_id on published projects and optimizes performance for projects endpoint

diogob:


db/migrate/20160113195732_copy_project_city_from_project_account.rb
```diff
@@ -0,0 +1,31 @@
...
+    WHERE pa.project_id = projects.id
+)
+WHERE
+    city_id IS NULL
+    AND EXISTS (
+        SELECT true
+        FROM project_accounts pa
+        WHERE pa.project_id = projects.id
+    )
+    AND projects.state_order >= 'published';
```
adrianob: Shouldn't this start from in_analysis since that is when the city becomes required?

db/migrate/20160113231834_migration_fix_missing_city_id.rb
```diff
@@ -0,0 +1,28 @@
...
+SET city_id = 1634
+WHERE
+    city_id IS NULL
+    AND projects.state_order >= 'published'
+    AND EXISTS (SELECT true FROM users u WHERE u.id = projects.user_id AND u.address_city = 'BH');
+
+-- Cases where the city is BH
+UPDATE projects
+SET city_id = 5274
+WHERE
```
adrianob: This where is missing ```AND EXISTS (SELECT true FROM users u WHERE u.id = projects.user_id AND u.address_city = 'SP');```

adrianob: Also we should do the same for RJ.

diogob: The SP is the default, so this is a catch-all for everything else with null city_id.
The remaining cases had empty cities, or cities made no sense (aaaaa, asdfg, etc)

db/migrate/20160113195732_copy_project_city_from_project_account.rb
diogob: Good catch! I'll change this.

db/migrate/20160113231834_migration_fix_missing_city_id.rb
diogob: The RJ cases apparently were fixed with either the project account or the user account

PR #2078, opened 2016-03-22
new publish page [#105306708]

adrianob:


app/controllers/flexible_projects_controller.rb
```diff
@@ -1,5 +1,7 @@
 #@TODO: this controller is total for legacy dashboard support
 # we will drop this soon when dashboard goes to API :)
+class InvalidProject < StandardError; end
```
devton: lets use ActiveRecord::Rollback instead StandardError

PR #2089, opened 2016-03-30
Small fixes

devton:


app/assets/javascripts/app/contribution/contribution_review.js
```diff
@@ -43,13 +43,28 @@ App.addChild('ReviewForm', _.extend({
   },
 
   makeFieldsRequired: function(){
+      console.log('foo');
```
vicnicius: debugger; x) let's remove it.

PR #2100, opened 2016-04-19
add direct message table [#115448127]

adrianob:


db/migrate/20160412140950_create_direct_messages_view.rb
```diff
@@ -0,0 +1,39 @@
...
+        SELECT dm.user_id, dm.to_user_id, dm.project_id, dm.from_email, dm.from_name, dm.content
+        from direct_messages dm;
+
+      CREATE OR REPLACE FUNCTION insert_direct_message() RETURNS trigger
+          LANGUAGE plpgsql
+          AS $$
+              declare
+                direct_message "1".direct_messages;
+              begin
+                insert into public.direct_messages (user_id, to_user_id, project_id, from_email, from_name, content) values (current_user_id(), NEW.to_user_id, NEW.project_id, NEW.from_email, NEW.from_name, NEW.content);
```
devton: @adrianob i think that we can add some kind of lock here when user try send many direct_messages, what you think?

PR #2130, opened 2016-05-02
update css

thiagocatarse:


thiagocatarse:
@vicnicius I removed the body-project.closed from extra and put it back into the _main

PR #2219, opened 2016-07-01
require login to report project [#123948611]

adrianob:


db/migrate/20160701172831_update_insert_project_report.rb
```diff
@@ -0,0 +1,27 @@
...
+  def change
+    execute <<-SQL
+      CREATE OR REPLACE FUNCTION insert_project_report() RETURNS trigger
+          LANGUAGE plpgsql
+          AS $$
+              declare
+                project_report "1".project_reports;
+              begin
+                insert into public.project_reports (user_id, project_id, email, reason, details, created_at, updated_at) values
+                                                  (NEW.user_id, NEW.project_id, NEW.email, NEW.reason, NEW.details, current_timestamp, current_timestamp);
```
devton: we should use current_user_id() since user is required.
revoke anon permissions too

PR #2221, opened 2016-07-04
Fixes reward input masks

vicnicius:


vicnicius:
Fixes reward input masks

PR #2251, opened 2016-07-28
remove old tables

adrianob:


db/migrate/20160728195033_clean_database.rb
```diff
@@ -0,0 +1,39 @@
+class CleanDatabase < ActiveRecord::Migration
+  def change
+    execute <<-SQL
+
+  CREATE  OR REPLACE FUNCTION state_order(project_id integer) RETURNS project_state_order
```
devton: @adrianob  we dont need to use case anymore

PR #2274, opened 2016-08-20
add is_follow to projects api

adrianob:


devton:
@adrianob why we are chaging from /assets/ to ../images? and build error NameError: uninitialized constant AddContributedByFriendsToProjects

PR #2286, opened 2016-08-31
Adds admin role check to data-user

vicnicius:


app/models/user.rb
```diff
@@ -270,7 +270,8 @@ def to_analytics
       last_login: self.last_sign_in_at,
       created_today: self.created_today?,
       follows_count: follows.count,
-      followers_count: followers.count
+      followers_count: followers.count,
+      is_admin_role: self.try(:admin?) || false
```
devton: can just use self.admin?

PR #2302, opened 2016-09-09
add project transitions endpoint

adrianob:


db/migrate/20160906153336_add_delete_projects_endpoint.rb
```diff
@@ -0,0 +1,23 @@
+class AddDeleteProjectsEndpoint < ActiveRecord::Migration
+  def change
+    execute <<-SQL
+      CREATE OR REPLACE FUNCTION "1".delete_project(_project_id integer) RETURNS void
+          LANGUAGE plpgsql
+          AS $$
+              begin
+                update project_transitions pt set most_recent = false where pt.project_id = _project_id;
```
devton: @adrianob we need to check permissions here before make updates.
- load the project inside a project var and check public.is_owner_or_admin(p.user_id)
- when project already have online and received some contributions or can delete only when current_state is draft?

db/migrate/20160906153336_add_delete_projects_endpoint.rb
```diff
@@ -0,0 +1,31 @@
...
+            begin
+                select * from public.projects where id = _project_id into v_project;
+
+                if _project_id is null or not public.is_owner_or_admin(v_project.user_id) or v_project.state <> 'draft' then
+                    raise exception 'invalid project permission';
+                end if;
+
+                update project_transitions pt set most_recent = false where pt.project_id = _project_id;
+                insert into public.project_transitions (to_state, metadata, sort_key, project_id, most_recent, created_at, updated_at) 
+                values ('deleted', '{"to_state":"deleted", "from_state":' || (select p.state from projects p where id = _project_id) || '}', 0, _project_id, true, current_timestamp, current_timestamp);
```
devton: dont need to use select p.state, v_project in scope

PR #2304, opened 2016-09-12
Added sitemap.xml index and sitemap to robots.txt

begrossi:


public/robots.txt
```diff
@@ -3,3 +3,13 @@
...
+
+User-agent: *
+Disallow: /dbhero/
+Disallow: /api_token
+Disallow: */admin
+Disallow: */new-admin
+Disallow: */projects/
+Disallow: */users/*/edit
+
+Sitemap: https://www.catarse.me/sitemap.xml
```
devton: @begrossi this is the right path? if yes wee need to add a route on config/routes.rb

PR #2322, opened 2016-09-22
Rails upgrading to 4.2.7

devton:


adrianob:
https://github.com/pluginaweek/state_machine/issues/343

adrianob:
@devton everything seems to be working now

PR #2329, opened 2016-09-29
added gateway balance oprations api mirror

devton:


vicnicius:
@ton the migration failed to run on our circleci environment.

PR #2330, opened 2016-09-30
cancel online project [#128108045]

adrianob:


devton:
@adrianob I think that we are missing the refund action when user cancel project :)

db/migrate/20160919134343_add_cancel_project.rb
```diff
@@ -0,0 +1,30 @@
...
+                select * from public.projects where id = _project_id into v_project;
+
+                if _project_id is null or not public.is_owner_or_admin(v_project.user_id) or v_project.state <> 'online' then
+                    raise exception 'invalid project permission';
+                end if;
+
+                update project_transitions pt set most_recent = false where pt.project_id = _project_id;
+                insert into public.project_transitions (to_state, metadata, sort_key, project_id, most_recent, created_at, updated_at) 
+                values ('failed', '{"to_state":"failed", "from_state":' || v_project.state || '}', 30, _project_id, true, current_timestamp, current_timestamp);
+                update projects set state = 'failed' where id = _project_id;
```
devton: We can create a contribution_notification to all paid payments on project

app/controllers/projects_controller.rb
```diff
@@ -123,6 +123,13 @@ def update
       }).run()
 
       redirect_to edit_project_path(@project, anchor: params[:anchor] || 'home')
+    elsif params[:cancel_project] == 'true'
+      api = ApiWrapper.new current_user
+      api.request("rpc/cancel_project", {
+        body: { _project_id: @project.id }.to_json,
+        action: :post
+      }).run()
```
devton: currently we dont have habilities to call direct refund from API, so maybe we can use something like:
```ruby
@project.refund_all_payments if @project.reload && @project.failed?
```
we also can notify users instead create a contribution_notification directly on RPC abstracting the refund and notify_users method on project_observer

PR #2350, opened 2016-10-17
New thank you

vicnicius:


db/migrate/20161014032607_add_recommendations_endpoint.rb
```diff
@@ -0,0 +1,67 @@
...
+                             recommendations_1.id AS project_id,
+                             0                    AS count
+                      FROM   contributions b
+                      JOIN   projects p
+                      ON     b.project_id = p.id
+                      JOIN   projects recommendations_1
+                      ON     recommendations_1.category_id = p.category_id
+                      WHERE  was_confirmed(b.*)
+                      AND    recommendations_1.state::text = 'online'::text) recommendations
+    WHERE    NOT (
```
devton: @vicnicius we should use ```public.is_owner_or_admin(b.user_id)``` to avoid users accessing another user recommendations.

PR #2390, opened 2016-11-09
Update project_success.html.slim

biabouskela:


app/views/catarse_bootstrap/user_notifier/mailer/project_success.html.slim
```diff
@@ -26,6 +26,6 @@ p
   | Um abraço apertado de parabéns!
 br/
 br/
-| Abraços
+| Abraços,
 br/
-| #{company_name}
+| Equipe do Catarse
```
devton: use ```Equipe do #{company_name}```

PR #2459, opened 2017-01-19
added en and pt translation

sushant12:


config/locales/catarse_bootstrap/views/layouts/footer.pt.yml
```diff
@@ -2,6 +2,9 @@ pt:
   layouts:
     footer:
       links:
+        how_it_works: Como funcion
```
devton: missing 'a' at end "Como funciona"

config/locales/catarse_bootstrap/views/layouts/footer.pt.yml
```diff
@@ -19,7 +22,6 @@ pt:
         github: Github
         instagram: Instagram
         near: Próximos a mim
-        press: Imprensa
```
devton: press is used

sushant12: I removed the duplicate `press` key .

PR #2491, opened 2017-02-13
Bellum

adrianob:


app/controllers/projects/contributions_controller.rb
```diff
@@ -87,6 +87,22 @@ def receipt
...
+    authorize resource
+    resource.delivery_status = 'received'
+    resource.save!
+    return render nothing: true
+  end
+
+  def update_status
+    project = Project.find params['project_id']
+    authorize project, :update?
+    Contribution.where(id: params['contributions']).update_all(delivery_status: params['delivery_status'])
```
devton: this should be project.contributions not?

db/migrate/20170210204756_add_deliver_fields_to_contribution.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeliverFieldsToContribution < ActiveRecord::Migration
+  def change
+    add_column :contributions, :delivery_status, :text, default: 'undelivered'
```
devton: what you think to add delivered_sent_at field too?

PR #2503, opened 2017-02-24
Adding more refresh views to stats views

begrossi:


sushant12:
Can you please tell me what is refresh views ?

adrianob:
@sushant12 take a look at the documentation for postgres materialized views. We use them for slow queries that don't need to be updated so often.

adrianob:
@sushant12 take a look at the documentation for postgres materialized views. We use them for slow queries that don't need to be updated so often.

sushant12:
Gotcha @adrianob . Thanks

PR #2554, opened 2017-03-27
add delivery notification

adrianob:


app/controllers/projects/contributions_controller.rb
```diff
@@ -105,10 +105,13 @@ def update_status
     project = Project.find params['project_id']
     authorize project, :update?
     contributions = project.contributions.where(id: params['contributions'])
-    contributions.update_all(delivery_status: params['delivery_status'])
     if params[:delivery_status] == 'delivered'
       contributions.update_all(reward_sent_at: Time.current)
+      contributions.where.not(delivery_status: 'delivered').each do |contribution|
```
devton: @adrianob  lets move this logic to another action in controller

app/views/catarse_bootstrap/user_notifier/mailer/delivery_confirmed.html.slim
```diff
@@ -0,0 +1,44 @@
...
+
+-if @notification.metadata.try(:[], 'message').empty?
+  p #{contribution.project.user.display_name} acabou de marcar que a sua recompensa foi ENVIADA.
+-else
+  p #{contribution.project.user.display_name} acabou de marcar que a sua recompensa foi ENVIADA e lhe mandou a seguinte mensagem:
+  p #{@notification.metadata.try(:[], 'message')}
+
+p Para te lembrar sobre esse apoio, enviamos as informações abaixo:
+p
+  | Nome do apoiador: #{contribution.user.display_name}
```
devton: we can use the _contribution_data partial on mailers templates

adrianob: In this case we don't want to show all the info that is in that partial.

PR #2557, opened 2017-03-28
add test pagarme key

adrianob:


config/initializers/pagarme.rb
```diff
@@ -1,5 +1,5 @@
 CatarsePagarme.configure do |config|
-  config.api_key = CatarseSettings.get_without_cache(:pagarme_api_key)
+  config.api_key = CatarseSettings.get_without_cache(Rails.env.production? ? :pagarme_api_key : :pagarme_test_api_key)
   config.ecr_key = CatarseSettings.get_without_cache(:pagarme_encryption_key)
```
devton: lets use the same logic for encryption key too

PR #2575, opened 2017-04-17
do not validate failed projects [#143839665]

adrianob:


devton:
@adrianob What happens when project is failed and user try edit and delete all data from inputs. i Think that we can put a accessor for going_fail or going_to_waiting for not validate on transition but continue validating when user trying to edit

adrianob:
@devton this is already only used for transitions, edit validation remains unchanged.

PR #2601, opened 2017-05-08
add public_tags endpoint

vicnicius:


db/migrate/20170508181435_adds_tags_endpoint.rb
```diff
@@ -0,0 +1,17 @@
+class AddsTagsEndpoint < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+    CREATE OR REPLACE VIEW "1".public_tags AS
+      SELECT
+          name,
+          slug
+      FROM public_tags;
+      GRANT SELECT ON "1".public_tags TO anonymous;
```
adrianob: We don't need permissions for the other roles?

vicnicius: I think that if we always make requests without token then we only the anonymous role would be enought, right? But just to make sure I added the others.

PR #2624, opened 2017-05-17
Using similarity on tag search

begrossi:


db/migrate/20170517202235_adds_tagsearch_endpoint.rb
```diff
@@ -0,0 +1,27 @@
+class AlterTagsEndpoint < ActiveRecord::Migration
```
devton: same name here

PR #2666, opened 2017-06-22
fix slip notification

adrianob:
Payments on create do not have gateway_data yet, resulting in a undefined method [] error.


devton:
@adrianob  specs failing

PR #2670, opened 2017-06-26
Adjusting dropdown toggle css

thiagocatarse:


thiagocatarse:
@vicnicius @adrianob you guys were right about the dropdown toggle classes.

The _main.scss wasn't updated. I hope now it is going to work vini

PR #2680, opened 2017-07-06
clean css

thiagocatarse:


app/assets/stylesheets/catarse_bootstrap/_main.scss
```diff
@@ -3640,47 +3501,6 @@ a:hover {
   background-color: #f1f4f4;
 }
 
-.hero-slide {
```
vicnicius: @thiagocatarse we still use this class on our hero.

PR #2760, opened 2017-08-28
Added routes to unsubscribe using a token

devton:


app/controllers/mail_marketing_users_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def subscribe
+    email = params['EMAIL']
+    mail_list = MailMarketingList.find_by(
+      list_id: (params[:list_id] || CatarseSettings[:sendgrid_newsletter_list_id])
+    )
+
+    if email =~ EMAIL_REGEX
+      user = User.find_by email: email
+      if user.present?
+        user.mail_marketing_users.create(
```
adrianob: Shouldn't we check if the user is already subscribed?

PR #2764, opened 2017-08-29
Mail marketing front

devton:


ACPK:
@adrianob - Is this ready to be pushed into master? I'm extremely excited about trying it out!

adrianob:
@ACPK not quite yet, it should be ready soon though.

PR #2771, opened 2017-09-04
Unsubscribe group

devton:
Get the marketing list suppressions group via api and delete marketing users that in suppression list


ACPK:
@vicnicius - Is this ready for master?

vicnicius:
@ACPK should be there already: https://github.com/catarse/catarse/commit/6af02fb112c2dc9d353bb5f5c0eb491da475a65f

PR #2783, opened 2017-09-18
removed unused code

sushant12:


vicnicius:
Thank you for the contribution, @sushant12! It's always good to remove stuff and we know we still have unused code in other places as well.

sushant12:
My contribution is like a grain of sand in the desert :p

PR #2840, opened 2017-11-13
initial common api front-end integration

vicnicius:


vicnicius:
Let's make sure we add the setting to all the environments when merged.

PR #2848, opened 2017-11-20
Process payment

vicnicius:
Quick additions to make payment process possible

### DIFF
--- a/app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
+++ b/app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
@@ -14,6 +14,7 @@ html
     meta name="root-url" content=root_url
     meta name="common-payment-api-host" content=CatarseSettings[:common_payment_service_api]
     meta name="common-analytics-api-host" content=CatarseSettings[:common_analytics_service_api]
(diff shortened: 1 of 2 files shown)


app/views/catarse_bootstrap/layouts/catarse_bootstrap.html.slim
```diff
@@ -14,6 +14,7 @@ html
     meta name="root-url" content=root_url
     meta name="common-payment-api-host" content=CatarseSettings[:common_payment_service_api]
     meta name="common-analytics-api-host" content=CatarseSettings[:common_analytics_service_api]
+    meta name="pagarme-encryption-key" content=CatarseSettings[:pagarme_encryption_key]
```
adrianob: Is this supposed to be public?

devton: yep

vicnicius: I also have mixed feelings about it also. But it looks like as soon as we don't provide the api_key, then it should be fine. 

https://docs.pagar.me/v2017-07-17/docs/api-key-e-encryption-key

PR #2856, opened 2017-11-27
Add common_id to user_details api

begrossi:


ACPK:
@begrossi - If I may ask, what is the common_id used for?

adrianob:
@ACPK that is the ID used in our new microservice architecture. This is mainly being used for subscription functionality for now but we intend to move all our code to this soon. More info available in these docs:
https://comum.gitbooks.io/docs/content/local-development-setup/setup.html

PR #2858, opened 2017-11-27
add common_id to some views that were missing it

vicnicius:


db/migrate/20171127171849_adds_common_id_to_endpoints.rb
```diff
@@ -0,0 +1,172 @@
+class AddsCommonIdToEndpoints < ActiveRecord::Migration
+  def change
+    execute <<-SQL
+    CREATE 
+    OR REPLACE VIEW "1"."user_details" AS       SELECT
```
adrianob: user_details was updated here https://github.com/catarse/catarse/commit/450750b617511b720cb1b90218cb9efde6e479a0

vicnicius: removed it. thx! o/

PR #2912, opened 2018-01-10
add sub report notify task

adrianob:


lib/tasks/cron.rake
```diff
@@ -188,6 +189,16 @@ namespace :cron do
     #end
   end
 
+  desc 'send subscription reports'
+  task notify_sub_reports: [:environment] do
+    SubscriptionProject.with_state(:online).each do |project|
```
devton: lets use find_each to avoid statement timeout

PR #2991, opened 2018-05-14
Update new_friends_contributions.html.slim

thiagocatarse:
Increasing the margin top at the main CTA button

### DIFF
--- a/app/views/catarse_bootstrap/user_notifier/mailer/follow_project_online.html.slim
+++ b/app/views/catarse_bootstrap/user_notifier/mailer/follow_project_online.html.slim
@@ -5,27 +5,33 @@
 
 center
   = image_tag follow.display_image, width: '50px', style: 'border-radius: 50px;'
-  h2 style="line-height: 5px"= follow.decorate.display_name
+  h2 style="line-height: 25px"= follow.decorate.display_name
   h3 style="font-weight: 500;"= "publicou um novo projeto"
 
 br/
 
-center
-  table style="border-spacing: 10px; width: 405px;"
-    tbody
-      tr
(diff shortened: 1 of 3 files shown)


app/views/catarse_bootstrap/user_notifier/mailer/follow_project_online.html.slim
```diff
@@ -5,27 +5,33 @@
...
+table border="0" cellpadding="0" cellspacing="0" style="min-width:100%;border-collapse:collapse" width="100%" 
+  tbody
+    tr
+      td style="padding-top:9px" valign="top" 
+        table align="left" border="0" cellpadding="0" cellspacing="0" style="max-width:300px;border-collapse:collapse" width="100%" 
+          tbody
+            tr
+              td style="padding-top:0;padding-left:18px;padding-bottom:9px;padding-right:18px;word-break:break-word;color:#3f4752;font-family:Helvetica;font-size:16px;line-height:150%;text-align:center" valign="top" 
+                div style="text-align:center" 
+                  = link_to(image_tag(project.display_image, width: '250px'), project_by_slug_url(project.permalink, {ref: 'notificacao_follow_published_project',utm_source:'notification',utm_medium:'email',utm_campaign:"follow_project_#{follow.id}",utm_content:project.permalink}))
```
thiagocatarse: I don't know if this is right. It would be nice to review. I know there are some inconsistencies on the way we've been using the link tag, but I didn't pay attention to this. I just would like to know if there is something wrong with the code (I kind of copied and pasted from another notification template that we use) @adrianob @grpse or @vicnicius 🤘

PR #3004, opened 2018-05-19
4 days or more boleto pending should not show on contributions report…

grpse:
… #156773495

Add new condition to project_contributions view


db/migrate/20180518204314_change_contributions_web_report.rb
```diff
@@ -0,0 +1,100 @@
+class ChangeContributionsWebReport < ActiveRecord::Migration
+  def up
+	execute %Q{
+REATE OR REPLACE VIEW "1"."project_contributions" AS
```
adrianob: typo here

adrianob:
@grpse there is still something wrong with the migration.

lib/tasks/cron.rake
```diff
@@ -206,6 +206,14 @@ namespace :cron do
     end
   end
 
+  desc 'Refuse boleto payments that are 4 days or more old and not paid'
+  task refuse_4_days_more_unpaid_boletos: [:environment] do
+    Payment.all_boleto_that_should_be_refused.each do |payment|
```
adrianob: Let's use ```find_each``` here instead.

PR #3027, opened 2018-06-08
taxes translation and projects total installments change #158173919

grpse:
Enable translation to taxes


db/migrate/20180608163915_change_max_installments_to_6.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeMaxInstallmentsTo6 < ActiveRecord::Migration
+  def change
+    execute <<-SQL
+      update projects set total_installments = 6 where total_installments = 3;
```
devton: wee need to add another migration to change default of proects total_installments

change_column_default :projects, :total_installments, 6

grpse:
default reset to 6 on total_installments

PR #3082, opened 2018-10-17
English Translations

zaidakram:
Added English Translations to be used by the rails as well as the i18n-js to be used in JS, for font-end application. I'll be opening up a PR at [catarse.js](https://github.com/catarse/catarse.js) as well for utilizing the locales better.

The translations of course are not perfect. But they are a good starting point for English version. As the English translations were left behind when `pt.yml` was broken off into multiple files under catarse_bootstrap folder.


adrianob:
Thanks a lot for this PR @zaidakram, amazing work!
For future reference, please use our new repo at https://github.com/common-group/services-core for all pull requests. Sorry we didn't put a warning in the README sooner.

PR #3084, opened 2018-11-07
README: Corrected versions of PostgRest & Ruby

Basit-qc:
Previous version of postgrest (0.3.0.3) does not support `limit` and `pagination` but we are using these parameters in new changes.
Also Ruby version is 2.4.4 in Gemfile, so corrected that as well.


VSPPedro:
Merge, please!

PR #3086, opened 2019-02-01
Update README.md

vsppedro:
Update ruby version in README to be the same as the version at the Gemfile


VSPPedro:
Sorry, I did not see that was already one Pull Request updating this. My bad.

